PR #6, opened 2017-11-13
RC1 Playlist UI

gilnoycloudinary:


src/assets/styles/components/playlist.scss
```diff
@@ -0,0 +1,217 @@
+@import 'compass/css3/flexbox';
+
+@mixin listLayout() {
+    .cld-p-layout {
```
mikeys: Modify classes namespace to `cld-plw` (short for playlist widget)`

src/assets/styles/main.scss
```diff
@@ -19,17 +19,26 @@ $icon-font-path: '../fonts' !default;
   line-height: 1.5em;
 }
 
+
+// Need to set base font size for em to all cloudinary components
+.cld-video-player,
```
mikeys: @gilnoycloudinary as a reminder - need to make sure that light skin works correctly.

src/components/playlist/panel/playlist-panel.js
```diff
@@ -0,0 +1,83 @@
...
+    if (this.options_.showAll) {
+      return playlist.list();
+    }
+
+    const startIndex = playlist.currentIndex() + 1;
+    const endIndex = startIndex + this.options_.total;
+    let selectedItems = playlist.list().slice(startIndex, endIndex);
+
+    if (selectedItems.length < this.options_.total) {
+      selectedItems = [...selectedItems, ...playlist.list().slice(0, this.options_.total - selectedItems.length)];
```
mikeys: Currently, for the case of total items of 6 and only 2 items in the playlist, the widget will show: 
item1, item2, item1, item2

which is incorrect. Consider using the following logic:


```javascript
        const items = [];

        const numOfItems = this.options_.total;

        while (items.length < numOfItems) {
          index = playlist.nextIndex(index);
          if (index === -1) {
            index = 0;
          }

          const source = playlist.list()[index];
          items.push(source);
        }

        return items;
      };
```

src/assets/styles/components/playlist.scss
```diff
@@ -0,0 +1,213 @@
...
+
+.cld-flex-col-2 {
+    flex: 0 1 50%;
+}
+
+.cld-flex-col-3 {
+    flex: 0 1 33.3%;
+}
+
+.cld-flex-col-4 {
```
mikeys: Remove 25% in case there are more than 4 items.

src/video-player.js
```diff
@@ -103,7 +111,7 @@ const overrideDefaultVideojsComponents = () => {
   children.push('titleBar');
   children.push('upcomingVideoOverlay');
   children.push('recommendationsOverlay');
-
+  // children.push('playlist');
```
mikeys: Remove commented out code?

src/video-player.js
```diff
@@ -214,24 +224,53 @@ class VideoPlayer extends Utils.mixin(Eventable) {
       }
     };
 
-    const _options = options.playerOptions;
-    const _vjs_options = options.videojsOptions;
+    let _playlistWidget;
 
-    console.log(_options, _vjs_options);
+    const initPlaylistWidget = () => {
+      this.videojs.on('playlistcreated', () => {
```
mikeys: `this.on('playlistcreated', () => {` should work fine as well.

src/video-player.js
```diff
@@ -214,24 +224,53 @@ class VideoPlayer extends Utils.mixin(Eventable) {
...
+        const plwOptions = _options.playlistWidget;
 
+        if (isObj(plwOptions)) {
+          if (_options.fluid) {
+            plwOptions.fluid = true;
```
mikeys: Need to handle a case where we changed `fluid` from `true` to `false` after player creation.

src/video-player.js
```diff
@@ -293,9 +334,21 @@ class VideoPlayer extends Utils.mixin(Eventable) {
...
   }
 
+  changeSkin(skin) {
+    if (skin === 'dark' || skin === 'light') {
+      const prevSkin = this.videojs.options().skin;
+
+      this.videojs.removeClass(`cld-video-player-skin-${prevSkin}`);
+      this.videojs.addClass(`cld-video-player-skin-${skin}`);
+      this.videojs.options({ skin: skin });
+      this.videojs.trigger('themechange', { currSkin: skin, prevSkin: prevSkin });
```
mikeys: `trigger('skinchanged')` ?

src/video-player.js
```diff
@@ -317,6 +370,11 @@ class VideoPlayer extends Utils.mixin(Eventable) {
     return this;
   }
 
+  playItem(item) {
```
mikeys: I would keep `playItem` inside `playlist` and not expose it outside in the VideoPlayer.

src/video-player.js
gilnoycloudinary: ill get rid of it for the mean time, lets deal with it when we will modify the css

src/assets/styles/components/playlist.scss
gilnoycloudinary: i will calc dynamically since the total elements in view have no limit

src/video-player.js
```diff
@@ -137,7 +136,6 @@ class VideoPlayer extends Utils.mixin(Eventable) {
 
     const onReady = () => {
       setExtendedEvents();
-      setCssClasses();
```
gilnoycloudinary: make skin method on player

PR #11, opened 2017-11-28
Bound context menu to Player container

mikeys:
- [X] Makes sure context menu does not go out of the player's bounds.

### DIFF
--- a/src/plugins/context-menu/components/context-menu.js
+++ b/src/plugins/context-menu/components/context-menu.js
@@ -1,5 +1,6 @@
 import videojs from 'video.js';
 import ContextMenuItem from './context-menu-item';
+import { setPosition } from '../../../utils/positioning';
 
 const Menu = videojs.getComponent('Menu');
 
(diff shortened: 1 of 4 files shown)


src/plugins/context-menu/components/context-menu.js
```diff
@@ -39,12 +39,20 @@ class ContextMenu extends Menu {
     });
   }
 
+  setPosition(left, top) {
```
gilnoycloudinary: i think we should add it to Util

src/plugins/context-menu/components/context-menu.js
```diff
@@ -39,12 +39,20 @@ class ContextMenu extends Menu {
...
+    this.el().style.left = `${left}px`;
+    this.el().style.top = `${top}px`;
```
gilnoycloudinary: i think we should add it to Util

mikeys: @gilnoycloudinary Done.

PR #23, opened 2018-03-08
CLD-5860

tsi:
This contains:
* Light playlist styles
* Playlist current item display

### DIFF
--- a/src/assets/styles/components/playlist.scss
+++ b/src/assets/styles/components/playlist.scss
@@ -109,7 +109,7 @@
     position: absolute;
     bottom: 0;
     left: 0;
-    padding: 9% 5% 4.5% 5%;
+    padding: 5% 0 5% 7%;
     width: 100%;
     transition: color .25s;
     @include display-flex(flex);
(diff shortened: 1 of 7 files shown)


src/assets/styles/components/playlist.scss
```diff
@@ -91,25 +91,25 @@
...
+.cld-plw-item-title-next {
     font-weight: bold;
 }
 
 .cld-plw-item-info-wrap {
     position: absolute;
     bottom: 0;
     left: 0;
-    padding: 10px;
+    padding: 5% 0 5% 7%;
```
gilnoycloudinary: @tsi  why %?

tsi: Nothing super critical, could just as well use pixels but since you asked :) -
a. to be consistent with the [recommendations styles](https://github.com/cloudinary/cloudinary-video-player/blob/master/src/components/recommendations-overlay/recommendations-overlay.scss#L127).
b. since box height is not constant but relative to player width, it makes sense to use a relative unit.

gilnoycloudinary: @tsi im not convinced, but we can stay with it for now

PR #26, opened 2018-03-16
remove save-dev dependency

lookfirst:
@mikeys typed `npm install save-dev foo` instead of `npm install --save-dev foo`

https://github.com/cloudinary/cloudinary-video-player/pull/13/files


lookfirst:
Mind you, some of the `devDependencies` are probably now in the `dependencies` section. You probably want to clean that up as well.

PR #28, opened 2018-03-19
Cld 6275 video request timeout

perelman:


src/video-player.js
```diff
@@ -144,6 +144,9 @@ class VideoPlayer extends Utils.mixin(Eventable) {
       if (source) {
         this.source(source);
       }
+      let maxTries = (options.videojsOptions.maxTries) ? options.videojsOptions.maxTries : 3;
```
mikeys: In JavaScript 0 value is false in conditions flow:
`0 ? 0 : 3`
`=> 3` 

That means that there's no way to set maxTries to 0.

mikeys: Also, perhaps a more elegant way to write the same would be:
`let maxTries = options.videojsOptions.maxTries || 3`

src/video-player.js
```diff
@@ -274,11 +278,32 @@ class VideoPlayer extends Utils.mixin(Eventable) {
...
+
       if (ready) {
         ready(this);
       }
     });
 
+
+    this.nbCalls = 0;
+    this.reTryVideo = (maxNumberOfCalls, timeout) => {
+      if (!this.isVideoReady()) {
```
mikeys: Based on the previous comment - the fact that we can modify the video player's source in any given time means that we need to make sure to reset the timers here on such event, otherwise weird behaviour may occur.

yaniv-li:
fixed comments

src/video-player.js
```diff
@@ -274,11 +275,36 @@ class VideoPlayer extends Utils.mixin(Eventable) {
...
+          let e = new Error('Video is not ready please try later');
+          this.videojs.trigger('error', e);
+        }
+      }
+
+    };
+
+    this.isVideoReady = () => {
+      let s = this.videojs.readyState();
+       if (s === 4) {
```
mikeys: @yaniv-li 
Unless I'm missing something this implementation still have a problem, given the following case:
1. Set source for the first video.
2. Video is struggling to load, and `nbCalls` has been increased. 
3. User decides not to wait and change the video source while waiting
The `nbCalls` variable will not get reset for the next source since readyState has not changed to 4 yet causing a shorter period for getting a timeout.

src/video-player.js
```diff
@@ -336,6 +362,10 @@ class VideoPlayer extends Utils.mixin(Eventable) {
     if (VideoPlayer.allowUsageReport()) {
       options.usageReport = true;
     }
+    this.nbCalls = 0;
```
mikeys: You should also use `clearTimeout` on the `retryVideo` timeout before reseting nbCalls. Otherwise, it may still affect `nbCalls` count for the next source you're setting for the video.

src/video-player.js
mikeys: Also, there's something wrong in the indentation here (`let` and `if` are not aligned).

src/video-player.js
```diff
@@ -336,6 +362,11 @@ class VideoPlayer extends Utils.mixin(Eventable) {
     if (VideoPlayer.allowUsageReport()) {
       options.usageReport = true;
     }
+    clearTimeout(this.reTryVideo);
+    this.nbCalls = 0;
+    let maxTries = this.videojs.options_.maxTries || 3;
+    let videoReadyTimeout = (this.videojs.options_.videoTimeout) ? this.videojs.options_.videoTimout : 55000;
```
mikeys: let videoReadyTimeout = this.videojs.options_.videoTimout || 55000

PR #39, opened 2018-04-29
Cld 7454 vast api improvements

yaniv-li:


src/video-player.js
```diff
@@ -287,6 +299,20 @@ class VideoPlayer extends Utils.mixin(Eventable) {
     });
 
 
+    if (Object.keys(options.playerOptions.ads).length > 0 && typeof this.videojs.ima === 'object') {
```
mikeys: In general looks good, but I assume you'd probably want some way for the user to actually turn this off on demand.

yaniv-li: if there is no ads option then it turns it off

mikeys: Once the player has been initialised, is there currently a way to turn it off?

yaniv-li: Sorry I don't follow

mikeys: Once you create a player instance with "ads" options to enable ads, is there currently a way to turn it off?

yaniv-li: no there is no option, it wasn't in the requirements

PR #40, opened 2018-04-30
muted ad issue

yaniv-li:
fixed ad shows as not muted when muted

### DIFF
--- a/src/plugins/ima/index.js
+++ b/src/plugins/ima/index.js
@@ -264,6 +264,10 @@
         this.settings.postrollTimeout);
 
       this.adsLoader.requestAds(adsRequest);
+      if (this.player.muted()) {
+
+        setAdMuted(true);
+      }
     }.bind(this);
 
     /**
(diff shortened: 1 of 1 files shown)


src/plugins/ima/index.js
```diff
@@ -264,6 +264,10 @@
         this.settings.postrollTimeout);
 
       this.adsLoader.requestAds(adsRequest);
+      if (this.player.muted()) {
```
mikeys: `setAdMuted(this.player.muted())`

src/plugins/ima/index.js
```diff
@@ -1035,6 +1023,25 @@
...
+        removeClass_(this.muteDiv, 'ima-non-muted');
+        this.adsManager.setVolume(0);
+        // Bubble down to content player
+        this.player.muted(true);
+        this.adMuted = true;
+        this.sliderLevelDiv.style.width = "0%";
+      } else {
+        addClass_(this.muteDiv, 'ima-non-muted');
+        removeClass_(this.muteDiv, 'ima-muted');
+        this.adsManager.setVolume(1);
```
mikeys: Shouldn't you use the current volume of the player rather then set volume to 100%?

src/plugins/ima/index.js
```diff
@@ -1035,6 +1020,25 @@
...
+        this.adMuted = true;
+        this.sliderLevelDiv.style.width = "0%";
+      } else {
+        addClass_(this.muteDiv, 'ima-non-muted');
+        removeClass_(this.muteDiv, 'ima-muted');
+        this.adsManager.setVolume(this.player.volume());
+        // Bubble down to content player
+        this.player.muted(false);
+        this.adMuted = false;
+        this.sliderLevelDiv.style.width = this.player.volume() * 100 + "%";
```
mikeys: I assume this line needs to be changed as well?

yaniv-li: way? this your code I just refactored it

mikeys: @yaniv-li It doesn't mean that it's 100% correct :)

mikeys: True, my bad.

PR #46, opened 2018-05-27
CLD-7899-present-Upcoming-option

tsi:
This fixes
image

in players without ads.
`ima` is not initialized so `getAdManager` doesn’t exist.
in this scenario `this.player().ima` is actually a function, this is why it throws the error


yaniv-li:
looks good

yaniv-li:
approved

PR #48, opened 2018-06-06
CLD-7906-titles-cut-off

tsi:


src/assets/styles/components/playlist.scss
```diff
@@ -27,13 +27,22 @@
 
                 .cld-plw-item-info-wrap {
                     left: 150px;
-                    padding: 20px 15px;
+                    padding: 15px 10px;
                     top: 0;
+                    max-width: calc(100% - 150px);
```
mikeys:
```
https://caniuse.com/#feat=calc
```

Let's make sure (with Amit) that we're ok with this.

tsi: Any news about this? I think `calc()` is safe enough these days, and even if not supported nothing terrible should happen in this case...

gilnoycloudinary: @tsi what about flex ?

tsi: @gilnoycloudinary can do without flex/calc

src/assets/styles/components/playlist.scss
```diff
@@ -27,13 +27,23 @@
 
                 .cld-plw-item-info-wrap {
                     left: 150px;
-                    padding: 20px 15px;
+                    padding: 15px 10px;
                     top: 0;
+                    right: 0;
+                    width: auto;
```
gilnoycloudinary: @tsi no flex or calc is needed ?

tsi: That's correct

mikeys:
@tsi Approved from my end. @gilnoycloudinary ?

src/assets/styles/components/playlist.scss
gilnoycloudinary: @tsi great, merge!

PR #51, opened 2018-06-13
Version release tooling

mikeys:
## Tooling for making release process a bit easier:
- `npm run deploy` for `stable` version release
- `npm run deploy -- edge` for `edge` version release

### Versioning:
Assuming `version: 1.0.3-edge.4`:
- `npm run deploy` will result a version bump to `1.0.3`
- `npm run deploy -- edge` will result a version bump to `1.0.3-edge.5`

Assuming `version: 1.0.3`:
- `npm run deploy` will result a version bump to `1.0.4`
- `npm run deploy -- edge` will result a version bump to `1.0.4-edge.0`

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
     "build-all": "npm run clean && npm run build && npm run build-light && npm run build-minified && npm run build-minified-light",
     "precommit": "eslint src",
     "preversion": "npm run build-all",
-    "deploy:edge": "npm version prerelease && npm publish --tag edge && git push origin master --follow-tags"
+    "version-prerelease": "npm version $(node -e \"version = require('./package.json').version; semver = require('semver'); console.log(semver.inc(version, 'prerelease', undefined, 'edge'))\")",
+    "deploy:edge": "npm run version-prerelease && npm publish --tag edge && git push origin master --follow-tags"
   },
   "author": "Cloudinary",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


tsi:
@mikeys nice!
few notes:
1. Couldn't you use `$npm_package_version` to get current version?
2. This approach means that `edge` versions come after their `stable` release, for example
if I'm on `1.0.4` and I run this, I'm being bumped to `1.0.5-edge.0`, is that the expected behavior? I would expect to get `1.0.4-edge.1`...
3. wouldn't it be easier to maintain these version scripts (I guess there might be another one for a stable release) in a JS file and call it from there?

mikeys:
@tsi is it indeed the case? Should we expect 1.0.4-edge.1 after 1.0.4 is released? I think it makes more sense to move to 1.0.5-edge.1... that’s how semantic versioning work

Regarding extracting it out to another file - sure. Makes sense.

Wasn’t aware of $npm_package_version. Sounds good!

tsi:
Sorry, brain was off, sure, for years I've been doing:
1.0.0-alpha.x
1.0.0-beta.x
1.0.0-rc.x
1.0.0
1.0.1-alpha.x
etc...

tsi:
I might be missing something but aren't you missing the `npm publish` (with or without `--tag edge`) ?
Also, I know it was my suggestion, but we might want to make `edge` the default, and `stable` optional, I'm afraid someone will push a new `stable` when they actually wanted `edge` ⚠️ 
Other than that I really (!) like this.

mikeys:
@tsi Good pointers. Added some fixes accordingly.

tsi:
That's a really nice deploy script :)
We'll be able to really test it only on our next deploy, but I'm merging since there's no real danger...

PR #83, opened 2019-03-30
Fix/ie11 compatibility

maoznir:


tsi:
From Yaron:

> I agree that we should making this optional. We already have a “light” and “full” packages - let’s add it to the “full”

@maoznir is that possible? do you have a better suggestion (I'm not sure everybody who uses `full` will want the IE11 support)?

maoznir:
@tsi I'll try to find a better solution. maybe 
1. replace uses of object.assign and array.prototype.includes
2. understand why ie has a syntax error without adding "ie":"11" to babelrc

tsi:
@maoznir your solution seems fine, I wouldn't spend too much time adapting the code to IE11.
We just need to figure out how to make it optional for users that don't want it, possible solutions:
a. create a separate script (AKA polyfill) that adds IE11 support.
b. create a special `flavor`, we can name it `compat` which is like the `full` but also adds IE support.
But if this doesn't worth it and Yaron wants to get this in, I don't mind merging this in as is, that's a product question.

PR #100, opened 2019-06-18
Test update

yaniv-li:


tsi:
Looks good, did some sanity, should get more testing after the next `edge` release

PR #117, opened 2019-08-28
fix testUrl doesn't work with cors

mayashavin:


mayashavin:
@yaniv-li As I explained to @maoznir HEAD request does not work on CORS. So no. I need this to be GET in order to work.

yaniv-li:
HEAD does work with CORS you can read here https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
Could you please explain what are you trying to achieve? maybe we could find a solution for your problem

mayashavin:
@yaniv-li You're right. Changed back to HEAD. Apparently Chrome disabled support for CORS for localhost server

yaniv-li:
@mayashavin what does `'Sec-Fetch-Mode': 'no-cors'` do?

mayashavin:
@yaniv-li this is to allow server to see the mode of the request. `No-cors` to enable CORS for this request. You can refer the documentation here: https://w3c.github.io/webappsec-fetch-metadata/

PR #141, opened 2020-02-12
Add dynamic dash plugin load

maoznir:
This is a crud suggestion of dynamically loading the dash plugin


maoznir:
Closed as this was an experiment

PR #148, opened 2020-04-21
Apps 2454 shoppable video sidebar

yaniv-li:


src/video-player.js
```diff
@@ -520,6 +522,9 @@ class VideoPlayer extends Utils.mixin(Eventable) {
     if (publicId instanceof VideoSource) {
       return this.videojs.cloudinary.source(publicId, options);
     }
+    if (this.videojs.shoppable) {
+      this.videojs.shoppable(this.videojs, options.shoppable);
+    }
```
tsi, line 527: @yaniv-li should this be wrapped with a build condition? in `/src/plugins/index.js` you wrap it.

src/plugins/cloudinary/models/video-source.js
```diff
@@ -172,6 +172,9 @@ class VideoSource extends BaseSource {
       if (codecTrans && !hasCodecSrcTrans) {
         opts.transformation.push(codecTrans);
       }
+      if (opts.format === 'auto') {
+        delete opts.format;
+      }
```
tsi, line 177: @yaniv-li does that belong here?

src/plugins/cld-interactive-video/index.js
```diff
@@ -0,0 +1,11 @@
+import ShoppableWidget from '../../components/shoppable-bar/shoppable-widget';
+
+export default function cldInteractivePlugin(player, options) {
+  if (options) {
+    // eslint-disable-next-line no-new
+    new ShoppableWidget(player, options);
+  }
+
+}
+
```
tsi, line 10: @yaniv-li is this plugin being used anywhere?

yaniv-li: this is the shoppable plugin

src/plugins/cloudinary/models/video-source.js
yaniv-li: this is the backend CDN chose best option

src/video-player.js
yaniv-li: it can be but it not really required as this checks if the shoppable plugin is loaded

PR #174, opened 2020-07-14
fix no MediaSource in ios

yaniv-li:


docs/scripts.js
```diff
@@ -26,14 +26,19 @@ function isIPaddress(ipaddress) {
...
+}
 // Get scripts & styles from:
 // `localhost` while developing
 // `unpkg.com` while demoing OR if a specific version is specified
 // These SHOULD be global since they are called by some examples
 var loadScript = function (source, ver) {
   var external = source.startsWith('http');
   var from =
-    !ver && (external || window.location.hostname === 'localhost' || isIPaddress(window.location.hostname))
+    !ver && (external || window.location.hostname === 'localhost' || isIPaddress(window.location.hostname) || isLocal())
```
maoznir, line 41: Please put all the logic in a single function, merged with the logic from isLocal

maoznir: I want future devs to understand what we check here by reading the function's name

yaniv-li: @maoznir I don't understand the comment this function just checks if this is coming from a local domain and this is the name of the function.

yaniv-li: I updated the comment to reflect that too

PR #175, opened 2020-07-16
removed dash from light version

yaniv-li:


tsi:
I don't like having multiple places that deal with the full/light switch. in the long term it can make our life harder.
See https://github.com/cloudinary/cloudinary-video-player/pull/176

PR #178, opened 2020-07-19
Add script for generating dev build of light version

maoznir:


maoznir:
@tsi I manually tested using docs/api.html with updated script tag to point to generated light.js

PR #183, opened 2020-08-09
Add support for setting source format in transformation

maoznir:


src/plugins/cloudinary/models/video-source.js
```diff
@@ -198,39 +198,42 @@ class VideoSource extends BaseSource {
...
-      let hasCodecSrcTrans = (isKeyInTransformation(opts.transformation, 'video_codec') || isKeyInTransformation(opts.transformation, 'streaming_profile'));
-      [type, codecTrans] = formatToMimeTypeAndTransformation(sourceType);
-      // If user's transformation include video_codec then don't add another video codec to transformation
-      if (codecTrans && !hasCodecSrcTrans) {
-        opts.transformation.push(codecTrans);
-      }
-      if (opts.format === 'auto') {
-        delete opts.format;
-      }
+  generateSource(sourceType) {
```
maoznir, line 201: I've refactored a bit

src/plugins/cloudinary/models/video-source.js
```diff
@@ -198,39 +198,42 @@ class VideoSource extends BaseSource {
...
+    // Create source transformation array
+    let transformation = this.sourceTransformation()[sourceType] || this.transformation();
+    transformation = Array.isArray(transformation) ? transformation : [transformation];
+    const sourceTypeFormat = normalizeFormat(sourceType);
+    const isAdaptive = (['mpd', 'm3u8'].indexOf(sourceTypeFormat) > -1);
+    const transformationFormat = (transformation.find(t => t && t.format) || {}).format;
+    // Get format from transformation, or from sourceType
```
maoznir, line 208: Here we allow to get format from transformation instead of from sourceType, but only for generating the url, isAdaptive is still determined by the sourceType and not by the transformation.

maoznir:
Closing, we've decided that to play videos with different file extension then the sourceType, use will pass full url.

PR #187, opened 2020-08-12
raw url support

yaniv-li:
Added support for playing raw urls as a public_id
if video does not have an extension you need to add a sourceType with the type
I added unit tests for this feature as well.
@tsi please look at the example page I added.


tsi:
@yaniv-li changed the example page to use `source()`
Works the same but made more sense to me in that context.

maoznir (approved):

PR #191, opened 2020-08-17
APPS-4025-test-pages

tsi:
This PR:
* Adds light/non-minified to test pages version-selector UI for easier testing
* Improves selector UI
* Adds `Video Transformations` test page
* Fix CSS loader so it will always pull the correct version (was pulling full instead of light in some edge cases)


tsi:
this branch is wrong. ignore. I will recreate.

PR #195, opened 2020-08-18
SNI-3020-update-logo

tsi:


tsi:
@maoznir this PR updates the player default logos as requested.
also updated the favicons on the example pages.
pls review

docs/360.html
```diff
@@ -3,7 +3,7 @@
 <head>
   <meta charset="utf-8">
   <title>Cloudinary Video Player</title>
-  <link href="https://cloudinary-res.cloudinary.com/image/asset/favicon-32x32-2991a47c2caa80211bf2dbf3f29317c8.png" rel="icon" sizes="32x32" type="image/png">
+  <link href="https://res.cloudinary.com/cloudinary-marketing/image/upload/f_auto,q_auto/c_scale,w_32/v1597183771/creative_staging/cloudinary_internal/Website/Brand%20Updates/Favicon/cloudinary_web_favicon_192x192.png" rel="icon" type="image/png">
```
maoznir, line 6: @tsi why remove the sizes attribute?

tsi: @maoznir because we don't provide the "other" sizes.
if we want we could provide the full list of icon/sizes like we do in other projects but it will add a lot of bloat into these example pages which will make them less readable, and IMO these specifically should be as simple/readable as possible.

maoznir: ok!

maoznir (approved):

PR #196, opened 2020-08-19
Replace usage of Array.find with utils/find

maoznir:


tsi (approved):
Code change looks good but needs testing, I'd list the places it touches so that QA can make sure it works (on IE, and in general)

tsi:
Here, fill free to correct if something wrong.
QA will need to verify:
* Shoppable hotspots
* Dash audio player
* Transformation array
@maoznir do we have a Jira ticket to put this in?

maoznir:
Thanks @tsi!

PR #209, opened 2020-10-07
Play raw url

yaniv-li:


maoznir (approved):

jackieros:
@maoznir Is 'raw URL' the terminology we want to use for users in the examples 'doc'? 
We don't use that concept elsewhere in Cloudinary documentation. Why not just call it a 'transformation URL'?   
(I realize it's an extension of the 'raw transformation' concept supported in SDKs, but I still think we should avoid inventing new terminology when not necessary)

@njb90 - Are you in the loop on this terminology?

PR #210, opened 2020-10-07
Apps 5658 secure=true by default

yaniv-li:


src/video-player.js
```diff
@@ -312,7 +312,7 @@ class VideoPlayer extends Utils.mixin(Eventable) {
     const initCloudinary = () => {
       const opts = _options.cloudinary;
       opts.chainTarget = this;
-
+      _options.cloudinary.cloudinaryConfig.config('secure', true);
```
maoznir, line 315: Adding this to defaults.js would be better no?

maoznir (changes requested):

src/video-player.js
yaniv-li: ok

test/unit/cloudinaryConfig.test.js.mv
```diff
@@ -0,0 +1,17 @@
...
+const cld = cloudinary.Cloudinary.new({ cloud_name: 'demo' });
+cloudinary.VideoPlayer = VideoPlayer;
+
+
+describe('secure true test', () => {
+  it('test force secure true', () => {
+    jest.useFakeTimers();
+    document.body.innerHTML = '<div><video id="test"/></div>';
+    let vp = new VideoPlayer('test', { hideContextMenu: true }, false);
+    console.log(vp);
```
maoznir, line 15: @yaniv-li can you please add a test that checks the source is https
and a test that checks that when passing secure=false to cloudinary, it is http?

maoznir (changes requested):

test/unit/cloudinaryConfig.test.js.mv
yaniv-li: @maoznir I added explicit secure true **in the player options** not in the cloudinary options  
there are 2 tests for default  and explicit false

maoznir (approved):

maoznir (changes requested):
cloudinaryConfig.test.js fails

PR #216, opened 2020-10-25
Dash quality

yaniv-li:


src/plugins/videojs-http-source-selector/plugin.js
```diff
@@ -0,0 +1,84 @@
+import videojs from 'video.js';
+
+import SourceMenuButton from './components/SourceMenuButton';
+import SourceMenuItem from './components/SourceMenuItem';
+import qualitySelector from '../../components/qualitySelector/qualitySelector';
```
tsi, line 5: @yaniv-li this seems to be un-used

src/plugins/videojs-http-source-selector/plugin.js
```diff
@@ -0,0 +1,84 @@
...
+    return false;
+  }
+
+  /**
+  *
+  * We have to wait for the manifest to load before we can scan renditions for resolutions/bitrates to populate selections
+  *
+  **/
+  player.on(['loadedmetadata'], function(e) {
+    let qualityLevels = player.qualityLevels();
```
tsi, line 44: @yaniv-li this seems to be un-used

src/components/qualitySelector/qualitySelector.js
```diff
@@ -98,6 +124,18 @@ const qualitySelector = {
     if (sourceMenuButton) {
       const qualityLevels = player.qualityLevels();
       if (qualityLevels && qualityLevels.length > 1) {
+        let levels = qualityLevels.levels_.filter((q) => q.enabled);
+        if (levels.length === 1) {
+          let idx = qualityLevels.levels_.findIndex(l => l.id === levels[0].id);
+          sourceMenuButton.children()[1].children()[idx].selected(true);
```
tsi, line 130: @yaniv-li following our previous discussion 
for marking selected option, see [what they did here](https://github.com/cloudinary/cloudinary-video-player/pull/216/files#diff-3bba335ded3dcda3c175fc8e6c2dd199bbaba7641b11ca9dd014ec0a2090c131R18):
```
let levels = this.player().qualityLevels();
levels[i].enabled = true;
```

src/video-player.js
```diff
@@ -360,7 +360,7 @@ class VideoPlayer extends Utils.mixin(Eventable) {
 
     this.initQualitySelector = () => {
       if (_options.qualitySelector !== false) {
-        this.videojs.httpSourceSelector({ default: 'high' });
+        this.videojs.httpSourceSelector({ default: 'Auto' });
```
tsi, line 363: @yaniv-li should this be 'auto' or 'Auto'? [see readme](https://www.npmjs.com/package/videojs-http-source-selector#auto)
either way, I don't see any change when playing with this value

tsi (commented):
@yaniv-li code looks good and I tested this locally and it works great! for both HLS & DASH

few notes though:
pls see linter warnings about `no-unused-vars`
do we really need both Component and Plugin? can we unite them both into the plugin?
Also, there's no need for the `videojs-` in the plugin name `videojs-http-source-selector` :)

src/components/qualitySelector/qualitySelector.js
yaniv-li: it should be enabled this is just to make sure the visual part is ok

src/video-player.js
yaniv-li: I think we should re-implement this according to what our design

maoznir: show be lower case i think

maoznir (approved):

PR #217, opened 2020-10-28
Dash quality

yaniv-li:
added polyfill for IE11


src/components/qualitySelector/qualitySelector.js
```diff
@@ -5,6 +5,55 @@ import '../../plugins/videojs-http-source-selector/plugin';
 
 import './quality-selector.scss';
 
+if (!Array.prototype.findIndex) {
```
maoznir, line 8: why not
import { findIndex } from 'utils/find';

maoznir (changes requested):

maoznir (approved):

PR #218, opened 2020-11-03
change check vtt to work in IE11

yaniv-li:
fixes missing subtitles in IE11


src/utils/cloudinary.js
```diff
@@ -2,6 +2,7 @@ import Promise from 'promise-polyfill';
 import fetchPF from 'fetch-ponyfill/build/fetch-browser';
 import { cloudinaryErrorsConverter } from '../plugins/cloudinary/common';
 import { find } from 'utils/find';
+import { assign } from 'utils/assign';
```
maoznir, line 5: Where is assign used?

maoznir (changes requested):

src/utils/cloudinary.js
yaniv-li: @maoznir removed it

maoznir (approved):

PR #222, opened 2020-11-25
side effects declarations

yaniv-li:


types/video-player-tests.ts
```diff
@@ -2,7 +2,8 @@
  * This file is used to validate the type spec. The code is not actually run.
  */
 
-import VideoPlayer from './video-player';
+import VideoPlayer, {Options} from './video-player';
```
maoznir, line 5: @yaniv-li 
Do you think we should add typescript as a dev dependency and add a script like
"test:types": "tsc"
so that we can validate our ts files?

yaniv-li: we can, but it is not a must in my opinion

PR #227, opened 2021-01-27
autoplay fallback

yaniv-li:


src/utils/cloudinary.js
```diff
@@ -40,8 +40,8 @@ const handleCldError = (that, options) => {
           goodSrcs.forEach(s => {
             s.try = true;
           });
+          that.videojs.autoplay(that.videojs.autoplay() || that.playWasCalled);
```
maoznir, line 43: What if autoplay = false, player.play was called and then player.pause was called and then player.source was called? it will set autoplay to true, which is not what we want. no?

src/video-player.js
```diff
@@ -674,6 +676,7 @@ class VideoPlayer extends Utils.mixin(Eventable) {
   }
 
   play() {
+    this.playWasCalled = true;
```
maoznir, line 679: Is there a better approach then keeping a state of if play was called?
We will have this.playWasCalled=true even if pause/stop where called after play, no?

maoznir (changes requested):

src/utils/cloudinary.js
yaniv-li: this is only relevant to the state before the video was started as it takes care of fallback.
I can't really see why someone would make all these calls.

src/video-player.js
yaniv-li: see my first comment

src/utils/cloudinary.js
maoznir: Take an interface similar to youtube as an example, you could play->pause->switch video, in such case you'll get autoplay even if autoplay is set to false. this might not be the wanted behavior.

PR #240, opened 2021-04-19
clear retry timeout on play

yaniv-li:


src/video-player.js
```diff
@@ -627,7 +630,7 @@ class VideoPlayer extends Utils.mixin(Eventable) {
     // #if (!process.env.WEBPACK_BUILD_LIGHT)
     this.initQualitySelector();
     // #endif
-    clearTimeout(this.reTryVideo);
+    clearTimeout(this.reTryId);
```
maoznir, line 633: @yaniv-li 
1. can you please add a description to this pr, explaining what it fixes and how the pr solves the issue?
2. Are we missing a clearTimeout inside this.reTryVideo() when video is ready?
3. Can you please add a test for the issue fixed in this pr?
Thanks!

maoznir (changes requested):

src/video-player.js
yaniv-li: @maoznir 
1. By convention this PR fixes the issue the branch is named after.
2. The `clearTimeout(this.reTryVideo)`; was a typo
3. There is no easy way to add a unit test for this only an integration test, do we want to add more integration tests?

maoznir (approved):

PR #241, opened 2021-04-20
fix: APPS-9253 video-player fix import problem

e1adn:
changing imports to be relative.


docs/transformations.html
```diff
@@ -57,28 +57,28 @@
     <h1>Cloudinary Video Player</h1>
     <h3 class="mb-4">Transformations Arrays</h3>
 
-    <h4 class="mb-2">via source()</h3>
+    <h4 class="mb-2">via source()</h4>
```
e1adn, line 60: wtf?

docs/transformations.html
```diff
@@ -35,15 +35,15 @@
...
-      var player2 = cld.videoPlayer('player-2', {
+      cld.videoPlayer('player-2', {
         transformation: [
           { width: 400, crop: 'limit' },
           { effect: 'blur:500' },
           { effect: 'saturation:-100' },
         ],
       });
 
-      var player3 = cld.videoPlayer('player-3');
```
e1adn, line 46: player2 and player3 does nothing

webpack/common.config.js
```diff
@@ -48,7 +49,6 @@ module.exports = {
 
   resolve: {
     extensions: ['.js', '.scss'],
-    modules: [path.resolve(__dirname, '../src'), 'node_modules'],
```
e1adn, line 51: why? I guess it was for the absolute import. Do not think it such a good iead.

PR #243, opened 2021-04-20
fix: APPS-9253 let to const

e1adn:
spacing and change some let to const


maoznir (approved):

maoznir (changes requested):
Please fix the error raised in travis

PR #244, opened 2021-04-21
fix big button disappear in IE11

yaniv-li:
Fixes an issue in IE 11 that the big play button disappears after ~60 sec`

### DIFF
--- a/src/video-player.js
+++ b/src/video-player.js
@@ -510,6 +510,11 @@ class VideoPlayer extends Utils.mixin(Eventable) {
       }
     });
 
+    this.videojs.on('canplaythrough', () => {
+      // clear retry timeout
+      this.videojs.clearTimeout(this.reTryId);
+    });
+
     this.videojs.ready(() => {
       onReady();
 


src/video-player.js
```diff
@@ -510,6 +510,11 @@ class VideoPlayer extends Utils.mixin(Eventable) {
       }
     });
 
+    this.videojs.on('canplaythrough', () => {
```
maoznir, line 513: @yaniv-li should this replace the fix done in the previous pr or is it an addition?
Shouldn't we just make sure the retry() function clears the timeout when needed instead of listening for events?

yaniv-li: @maoznir 
I think it can replace the fix from the previous PR that said they are not conflicting.
I think listening to events is better than trying to figure out the state of the file after 55 seconds.

src/video-player.js
```diff
@@ -510,6 +510,11 @@ class VideoPlayer extends Utils.mixin(Eventable) {
...
+      // clear retry timeout
```
e1adn, line 514: In my opinion, the comment can be removed.. the function named 'clearTimeout' is clean enough

maoznir (approved):

e1adn (approved):

PR #245, opened 2021-04-21
Apps 9253 doc sub folders

e1adn:
move html files to sub folder and moving js from html to  js file


.eslintrc.js
```diff
@@ -6,6 +6,15 @@ module.exports = {
     "es6": true,
     "jest": true
   },
+  "overrides":[
+    {
+      "files": [ "docs/**/*.js" ],
```
e1adn, line 11: the doc files are not complied from es6 to es5, allowing using var

PR #252, opened 2021-05-26
Apps 5965 handle non existent texttracks

yaniv-li:
added unit test


test/unit/cloudinaryUtils.test.js
```diff
@@ -21,35 +21,33 @@ const tracks = [
...
-    jest.useFakeTimers();
+  it('test filter out bad vtt', async (done) => {
     let vjs = jest.createMockFromModule('video.js');
     vjs.addRemoteTextTrack = jest.fn();
     jest.spyOn(vjs, 'addRemoteTextTrack');
-    let a = await Utils.filterAndAddTextTracks(tracks, vjs);
-    console.log(a);
-    // expect(vjs.addRemoteTextTrack).toBeCalled();
-
+    filterAndAddTextTracks(tracks, vjs);
```
e1adn, line 46: Looks like 'filterAndAddTextTracks' is a function that make async action. 
It used fetch inside a forEach. But the function does not return anything.  
Using a  'setTimeout' is not a solution.  
Please make the function into a pure function.  it should  return a Promise 
 and resolve it when all the fetching are done.

e1adn: Or using Promise.All and call ' videojs.addRemoteTextTrack' after all fetch is done

yaniv-li: @e1adn I don't what to change working code just to satisfy a test method.
It is like that because these are basically "async workers" they do something we do not need return, as it would block the flow for no good reason, we use the work result in a later stage.  they build the result independent of each other.

PR #254, opened 2021-05-31
APPS-9490 adding option for adding trackers area

e1adn:
This feature is a first step for 'Video Zoom'. 

 There is a 'interaction-area' doc page.  (http://localhost:3000/interaction-area.html)
There isn't a link for that on the main page, customers should not have access to it yet.

https://user-images.githubusercontent.com/76470570/120180765-d8f4fa00-c214-11eb-9200-bff5bdcd60e1.mov

### DIFF
--- a/src/video-player.js
+++ b/src/video-player.js
@@ -10,8 +10,14 @@ import PlaylistWidget from './components/playlist/playlist-widget';
 // #if (!process.env.WEBPACK_BUILD_LIGHT)
 import qualitySelector from './components/qualitySelector/qualitySelector.js';
 // #endif
-
 import VideoSource from './plugins/cloudinary/models/video-source';
+import { createElement } from './utils/dom';
+import {
+  getMetaDataTracker,
+  getTrackerItem,
+  setTrackersContainer,
+  TRACKERS_CONTAINER_CLASS_NAME
+} from './video-player.utils';
 
 const CLOUDINARY_PARAMS = [
   'cloudinaryConfig',
@@ -168,6 +174,7 @@ overrideDefaultVideojsComponents();
 let _allowUsageReport = true;
 
 class VideoPlayer extends Utils.mixin(Eventable) {
+
   constructor(elem, options, ready) {
     super();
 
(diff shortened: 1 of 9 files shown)


src/video-player.js
```diff
@@ -606,6 +619,65 @@ class VideoPlayer extends Utils.mixin(Eventable) {
...
+        trackersOptions && trackersOptions.onClick({ item, index, event });
+      });
+    });
+
+    const tracksContainer = createElement('div', { 'class': TRACKERS_CONTAINER_CLASS_NAME }, trackerItems);
+
+    setTrackersContainer(this.videojs, tracksContainer);
+  }
+
+  addCueListener() {
```
e1adn, line 661: ״הכנה למזגן״

maoznir: This is why you removed this rule right?
"no-unused-expressions": "off"
why is the unused code part of this pr?

src/video-player.js
```diff
@@ -168,9 +174,12 @@ overrideDefaultVideojsComponents();
 let _allowUsageReport = true;
 
 class VideoPlayer extends Utils.mixin(Eventable) {
+
   constructor(elem, options, ready) {
     super();
 
+    this.isZoomed = false;
```
maoznir, line 181: Isn't there an object that contains all default configuration? better use such an object.

src/video-player.js
```diff
@@ -606,6 +619,65 @@ class VideoPlayer extends Utils.mixin(Eventable) {
...
+
+    this._addTrackersItems(tracksData, trackersOptions);
+  }
+
+  _addTrackersItems(trackersData, trackersOptions) {
+    const trackerItems = trackersData.map((item, index) => {
+      return getTrackerItem(item, (event) => {
+        this.isZoomed = true;
+        this._setGoBackButton();
+        trackersOptions && trackersOptions.onClick({ item, index, event });
```
e1adn, line 652: "no-unused-expressions": "off" for this

src/video-player.js
```diff
@@ -606,6 +619,65 @@ class VideoPlayer extends Utils.mixin(Eventable) {
     return this.videojs.cloudinary.cloudinaryConfig(config);
   }
 
+  _setGoBackButton() {
+    const button = createElement('button', { 'class': 'go-back-button' }, 'Go back');
+
+    button.addEventListener('click', () => {
+      this.goBack && this.goBack();
```
e1adn, line 626: "no-unused-expressions": "off" for this

src/video-player.js
e1adn: It is not a configuration, it is a private variable. i will change it to '_ isZoomed' and if it will need to be public in the future we can add a getter function (to prevent changing it out of the class)

src/video-player.js
```diff
@@ -606,6 +619,65 @@ class VideoPlayer extends Utils.mixin(Eventable) {
...
+    }, false);
+
+    const tracksContainer = createElement('div', { 'class': TRACKERS_CONTAINER_CLASS_NAME }, button);
+    setTrackersContainer(this.videojs, tracksContainer);
+  }
+
+  addTrackers(tracksData, trackersOptions) {
+    const currentSourceUrl = this.currentSourceUrl();
+
+    this.goBack = () => {
```
e1adn, line 636: I will change it from 'goBack' to 'unZoom'

maoznir (approved):

PR #259, opened 2021-06-08
Refactoring

e1adn:
some refactoring


maoznir (commented):
@e1adn looks good to me, I'd only suggest to document each function you move/add (what it does, input/output)

maoznir (approved):

PR #261, opened 2021-06-09
Apps 9676 add support for custome vtt

e1adn:
https://cloudinary.atlassian.net/browse/APPS-9674

You may take a look on 'http://localhost:3000/interaction-area.html' when running local dev

### DIFF
--- a/docs/interaction-area.html
+++ b/docs/interaction-area.html
@@ -157,7 +157,7 @@
          bottomSource: 'https://res.cloudinary.com/eran2903/video/upload/w_540,x_1260,y_3100,c_crop/w_270,eo_8,q_auto:best,f_auto/production_ID_3888264_fuuo7c.mp4'
       }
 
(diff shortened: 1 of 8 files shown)


docs/interaction-area.html
```diff
@@ -22,10 +22,6 @@
       var cld = cloudinary.Cloudinary.new({cloud_name: 'demo'});
       var options = {fluid: true};
 
-      player = cld.videoPlayer('player', options);
-
-      player.source('https://res.cloudinary.com/eran2903/video/upload/eo_8,w_270,q_auto:best,f_auto/production_ID_3888264_fuuo7c.mp4');
-
       var interactionAreas = [
```
maoznir, line 25: how about we add identifiers to each interaction are so that we can recognize it by name/id instead/in addition to index?

e1adn: will do

maoznir (approved):

PR #271, opened 2021-06-23
fix big button z-index bug

e1adn:


yaniv-li:
@tsi please take a look

yaniv-li (approved):

PR #273, opened 2021-06-29
APPS-9877 option { controlBar: false } generates errors

e1adn:
https://cloudinary.atlassian.net/browse/APPS-9877


tsi:
@e1adn do we really need to even [initialize the qualitySelector plugin](https://github.com/cloudinary/cloudinary-video-player/blob/edge/src/video-player.js#L320) if there are no controls?

tsi (approved):
this change should fix the errors.
but I don't know about this re-scoping, it will need QA -
image

PR #274, opened 2021-06-29
fix source copy

yaniv-li:


src/plugins/dash/videojs-dash.js
```diff
@@ -298,7 +299,7 @@ Html5DashJS.hooks_ = {};
 
 const canHandleKeySystems = function(source) {
   // copy the source
-  source = JSON.parse(JSON.stringify(source));
+  source = assign({}, source);
```
e1adn, line 302: Are you sure that changing from deep-copy to shallow-copy is enough? it will keep nested properties by reference

yaniv-li: it fixes the IE issue and I didn't see an issue with other browsers after the change the problem is that there are circular references JSON.stringify in IE can not handle.

e1adn (approved):

PR #277, opened 2021-07-07
Apps 9963 check if we can update dependencies

e1adn:
Before : 
Screen Shot 2021-07-07 at 11 26 07

After: 
Screen Shot 2021-07-07 at 11 24 20


e1adn:
'videojs-ima' has not been updated because it also need a node version update to do so

maoznir (approved):

PR #278, opened 2021-07-11
APPS-9923 - Do not add video codec if it already defined

e1adn:
APPS-9923 - Do not add video codec if it already defined


src/plugins/cloudinary/models/video-source/video-source.utils.js
```diff
@@ -28,3 +30,31 @@ export function normalizeFormat(format) {
 
   return res;
 }
+
+const isContainCodec = (value) => Object.values(VIDEO_CODEC).some((codec) => value.includes(codec));
+
+export const isCodecAlreadyExist = (transformations, rawTransformation) => {
+  if (!transformations) {
+    return false;
```
maoznir, line 38: what if the codec is in the raw transformation?

src/plugins/cloudinary/common.js
```diff
@@ -178,17 +178,25 @@ const h264avcToString = (s) => {
...
-    case 'hev1':
-      return { video_codec: 'h265' };
-    case 'h264':
-      return { video_codec: 'h264:baseline:3.0' };
+    case VIDEO_CODEC.VP9:
+      return { video_codec: VIDEO_CODEC.VP9 };
+    case VIDEO_CODEC.HEV1:
+      return { video_codec: VIDEO_CODEC.H265 };
+    case VIDEO_CODEC.H264:
+      return { video_codec: `${VIDEO_CODEC}:baseline:3.0` };
```
maoznir, line 199: missing .H264 ? 
if it really is missing, then we're missing a test for this

maoznir: Also, this is very verbose and surly can be refactored to something like:
if (isCodec(...)){
  return {video_codec: codec}
}

src/plugins/cloudinary/models/video-source/video-source.utils.js
```diff
@@ -28,3 +30,31 @@ export function normalizeFormat(format) {
 
   return res;
 }
+
+const isContainCodec = (value) => Object.values(VIDEO_CODEC).some((codec) => value.includes(codec));
```
maoznir, line 34: please make sure there are passing unit tests for the utils

maoznir: You should also add an integration test for the specific case this pr is handling

maoznir (changes requested):

test/unit/videoSource.test.js
```diff
@@ -297,3 +298,66 @@ describe('tests withCredentials', () => {
 });
 
 
+describe('test isCodecAlreadyExist method', () => {
```
maoznir, line 301: i'd also add tests for the generated url, to see that codec isn't added or is added when needed

maoznir (changes requested):

maoznir (approved):

test/unit/videoSource.test.js
e1adn: @maoznir done

PR #283, opened 2021-07-19
Video player - in case not clicking on 'Got it' button the screen sho…

e1adn:
https://cloudinary.atlassian.net/browse/APPS-10076

the 'do show It Again Checkbox' should be hidden by default

### DIFF
--- a/src/components/interaction-area/interaction-area.utils.js
+++ b/src/components/interaction-area/interaction-area.utils.js
@@ -91,7 +91,13 @@ export const shouldShowAreaLayoutMessage = (interactionLayoutConfig) => {
   return (!interactionLayoutConfig || interactionLayoutConfig.enable) && localStorage.getItem(INTERACTION_AREA_LAYOUT_LOCAL_STORAGE_NAME) !== 'true';
 };
 
-export const createInteractionAreaLayoutMessage = (videojs, onClick) => {
+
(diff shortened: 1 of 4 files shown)


src/components/interaction-area/interaction-area.utils.js
```diff
@@ -91,65 +91,60 @@ export const shouldShowAreaLayoutMessage = (interactionLayoutConfig) => {
   return (!interactionLayoutConfig || interactionLayoutConfig.enable) && localStorage.getItem(INTERACTION_AREA_LAYOUT_LOCAL_STORAGE_NAME) !== 'true';
 };
 
-export const createInteractionAreaLayoutMessage = (videojs, onClick) => {
+
+const onClickInteractionAreaLayoutClick = (checked, onClick) => {
+  localStorage.setItem(INTERACTION_AREA_LAYOUT_LOCAL_STORAGE_NAME, JSON.parse(checked));
+  onClick();
```
maoznir, line 97: if onClick might be null, add a check

maoznir (changes requested):

src/components/interaction-area/interaction-area.utils.js
e1adn: i will set noop as default :

`
const onClickInteractionAreaLayoutClick = (checked, onClick= noop) => {
`

maoznir (approved):

PR #285, opened 2021-07-21
prevent css glitch for interaction-areas-container

e1adn:
This is the jira ticket : https://cloudinary.atlassian.net/browse/APPS-10140.

The problem is since the elements are been added to the dom dynamically there is a small delay time till the css has been implemented.

### DIFF
--- a/src/components/interaction-area/interaction-area.utils.js
+++ b/src/components/interaction-area/interaction-area.utils.js
@@ -90,11 +90,18 @@ export const getZoomTransformation = (videoElement, interactionAreaItem) => {
 export const setInteractionAreasContainer = (videojs, newInteractionAreasContainer) => {
   const currentInteractionAreasContainer = getInteractionAreasContainer(videojs);
 
+  newInteractionAreasContainer.style.visibility = 'hidden';
+
   if (currentInteractionAreasContainer) {
     currentInteractionAreasContainer.replaceWith(newInteractionAreasContainer);
   } else {
     videojs.el().append(newInteractionAreasContainer);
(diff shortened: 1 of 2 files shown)


src/components/interaction-area/interaction-area.utils.js
```diff
@@ -90,11 +90,18 @@ export const getZoomTransformation = (videoElement, interactionAreaItem) => {
 export const setInteractionAreasContainer = (videojs, newInteractionAreasContainer) => {
   const currentInteractionAreasContainer = getInteractionAreasContainer(videojs);
 
+  newInteractionAreasContainer.style.visibility = 'hidden';
```
maoznir, line 93: I'm unfamiliar with what you're trying to do here, but from the description it sounds like hiding the current area is supposed to be the solution no?

maoznir (changes requested):

src/components/interaction-area/interaction-area.utils.js
e1adn: Ok... so after debug it I found out what causes the glitch .. the image takes some time to get loaded so the title is jumping ... the container gets replace with a new element so there is not state there is more than one in the dom.

As you can see in the picture I gave the image a height to fix the glitch.

Screen Shot 2021-07-22 at 7 49 44

maoznir (approved):

src/components/interaction-area/interaction-area.utils.js
maoznir: great

PR #292, opened 2021-08-03
APPS-10246 - Add validators for video player configs

e1adn:
Add validators, the be sure the configuration object is valid. 

https://cloudinary.atlassian.net/browse/APPS-10246

### DIFF
--- a/src/validators/validators-functions.js
+++ b/src/validators/validators-functions.js
@@ -3,15 +3,17 @@ import { castArray } from '../utils/array';
 
 const getValidators = (validator) => isFunction(validator) ? validator() : validator;
 
+/**
+ * check if a value is valid or not
+ * @param  {object} validators -  a config object
+ * @param  {any} value
+ * @param  {key} string
+ * @returns boolean - using the validators to check if the value is a valid value or not
+ */
 export const isValueValid = (validators, value, key) => {
-  let invalidItem = null;
-
-  const isValid = castArray(validators).some((validator) => {
-    const validatorItem = getValidators(validator);
-    const isValid = validatorItem.value(value);
-    invalidItem = isValid ? null : validatorItem;
-    return isValid;
-  });
+  const validatorsArray = castArray(validators);
+  const isValid = validatorsArray.some((validator) => getValidators(validator).value(value));
+  const invalidItem = validatorsArray.find((validator) => !getValidators(validator).value(value));
 
   if (invalidItem) {
     const validatorItem = getValidators(invalidItem);
@@ -21,6 +23,12 @@ export const isValueValid = (validators, value, key) => {
   return isValid;
 };
 
+/**
+ * check if a configuration object is valid or not
+ * @param  {object} config -  a config object
+ * @param  {object} validators
+ * @returns boolean - true is the configuration object is valid and false if it is not
+ */
 export const isValidConfig = (config, validators) => {
   if (isPlainObject(validators)) {
     for (let key in config) {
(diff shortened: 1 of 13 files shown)


src/validators/validators-functions.js
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * check if a value is valid or not
+ * @param  {object} validators -  a config object
+ * @param  {any} value
+ * @param  {key} string
+ * @returns boolean - using the validators to check if the value is a valid value or not
+ */
+export const isValueValid = (validators, value, key) => {
+  const validatorsArray = castArray(validators);
+  const isValid = validatorsArray.some((validator) => getValidators(validator).value(value));
```
maoznir, line 15: This should be validatorsArray.every instead of .some right?

src/validators/validators-functions.js
```diff
@@ -0,0 +1,50 @@
...
+  const invalidItem = validatorsArray.find((validator) => !getValidators(validator).value(value));
```
maoznir, line 16: Why is the isValid needed if you have the invalid check? can leave only the invalid check, no?

src/validators/validators.js
```diff
@@ -0,0 +1,107 @@
...
+      fontFace: validator.isString,
+      posterOptions: {
+        publicId: validator.isString,
+        transformation: [validator.isObject, validator.isArray]
+      }
+    }
+  }
+};
+
+export const sourceValidators = {
```
maoznir, line 75: nice!

maoznir (changes requested):

src/validators/validators-functions.js
e1adn: It's 'or' not an 'and' ... if at list one is valid it defined as valid

e1adn: for example, a value that can be a number or and string

src/validators/validators-functions.js
e1adn: it does not work and I understand why.... for example:  if the first validator is false the find will done (!getValidators) i need to find what causes a failure. but it does not say its in isValid because the next item can be true ... i will fix it

e1adn: @maoznir replace with a single forEach

src/validators/validators-functions.js
```diff
@@ -0,0 +1,56 @@
...
+    isValid = isItemValid && !isValid ? isItemValid : isValid;
+    invalidItem = !isItemValid && !invalidItem ? validator : invalidItem;
+  });
+
+  if (invalidItem && !isValid) {
+    const validatorItem = getValidators(invalidItem);
+    console.error(`cloudinary video player: ${validatorItem.message(key)}`);
+  }
+
+  return isValid;
```
maoznir, line 29: I suggest to improve these:
1. explain what 'key' is and/or name it better
2. explain that this function returns true if at least one of the validators returns true and otherwise it throws, (which is weird because you'd expect it to return true if all validators return true, but i'm not familiar with the spec/need here)
3. split this function's concerns into functions that do a single job/concern OR write it in a declarative manner

Plus a question: why do you need an array of validator if you only care that one returns true? i don't see a test for an array of validators, are you sure its needed?

maoznir (changes requested):

src/validators/validators-functions.js
e1adn: I removed the need for a validators array. The only property that had an array was 'transformation' (can be both object and array) I do not think we should check it since it is a SDK configuration and not a 'video player' configuration.  I rename 'isObject' to 'isPlainObject' (true only for {} ) and change 'isObject' to be true for both object and array

PR #300, opened 2021-08-12
Apps 10327 interaction areas as config and not a function 

e1adn:
https://cloudinary.atlassian.net/browse/APPS-10316

### DIFF
--- a/src/components/interaction-area/interaction-area.service.js
+++ b/src/components/interaction-area/interaction-area.service.js
@@ -28,7 +28,6 @@ export const interactionAreaService = (player, playerOptions, videojsOptions) =>
   let currentTrack = null;
   let unZoom = noop;
 
-  const shouldSetResize = () => isInteractionAreasEnabled();
   const shouldLayoutMessage = () => shouldShowAreaLayoutMessage(videojsOptions.interactionAreas);
 
   function isInteractionAreasEnabled(enabled = false) {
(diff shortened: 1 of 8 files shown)


src/components/interaction-area/interaction-area.service.js
```diff
@@ -25,35 +25,35 @@ export const interactionAreaService = (player, playerOptions, videojsOptions) =>
 
   let firstPlayed = false;
   let isZoomed = false;
-  let setStaticInteractionAreas = null;
   let currentTrack = null;
   let unZoom = noop;
 
-  const shouldSetResize = () => isInteractionAreasEnabled(setStaticInteractionAreas);
+  const shouldSetResize = () => isInteractionAreasEnabled();
```
maoznir, line 31: why is this variable needed?

src/components/interaction-area/interaction-area.service.js
```diff
@@ -25,35 +25,35 @@ export const interactionAreaService = (player, playerOptions, videojsOptions) =>
...
   }
 
   function updateTrack() {
     currentTrack && player.videojs.removeRemoteTextTrack(currentTrack);
 
-    if (!isInteractionAreasEnabled()) {
-      return;
-    }
-
+    const isEnabled = isInteractionAreasEnabled();
```
maoznir, line 42: why is this variable needed? the function name is good enough, no?
There is another place in the code where you use the function directly

maoznir (changes requested):

src/components/interaction-area/interaction-area.service.js
e1adn: I will change the function name to something more meaningful. 

I totally agree with you that we should check the condition of a need function outside of the function in case it is **pure function**   - A function that only gets some inputs and returns values and does not change anything on the parent scope. But it's not the case, the function does not return anything any only call other functions depend on the function logic.
In my opinion, it should be there.

src/components/interaction-area/interaction-area.service.js
e1adn: done

maoznir (approved):

PR #319, opened 2021-12-02
APPS-11107 - escape video tag id before passing to querySelector

aleksandar-cloudinary:
https://cloudinary.slack.com/archives/C9B1DMFV3/p1638468005034600

### DIFF
--- a/src/video-player.utils.js
+++ b/src/video-player.utils.js
@@ -9,6 +9,7 @@ import {
   AUTO_PLAY_MODE
 } from './video-player.const';
 import { isString } from './utils/type-inference';
+require('css.escape');
 
 export const addMetadataTrack = (videoJs, vttSource) => {
   return videoJs.addRemoteTextTrack({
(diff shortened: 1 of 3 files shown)


src/video-player.utils.js
```diff
@@ -9,6 +9,7 @@ import {
   AUTO_PLAY_MODE
 } from './video-player.const';
 import { isString } from './utils/type-inference';
+require('css.escape');
```
e1adn, line 12: Please change it to:
`import cssEscape from 'css.escape';`

src/video-player.utils.js
```diff
@@ -33,7 +34,7 @@ export const getResolveVideoElement = (elem) => {
     }
 
     try {
-      elem = document.querySelector(`#${id}`) || videojs.getPlayer(id);
+      elem = document.querySelector(`#${CSS.escape(id)}`) || videojs.getPlayer(id);
```
e1adn, line 37: and use the imported function : 
`cssEscape (id)`

src/video-player.utils.js
e1adn: Also please add a comment about it, something like:  'includes a polyfill to support ie11'

aleksandar-cloudinary: Done

e1adn (approved):

PR #325, opened 2021-12-22
APPS-11327-Upgrade-DashJS

tsi:
This removes our implementation of `videojs-contrib-dash` and uses the official one.


e1adn:
@tsi it would be nice also take care of that:
Screen Shot 2021-12-22 at 15 48 02

PR #327, opened 2021-12-22
APPS-11327-Upgrade-DashJS

tsi:
Replacing #325
This time keeping the internal implementation of `videojs-contrib-dash`.
Importing it as a module results in 2 copies of DashJS in the final bundle :(

### DIFF
--- a/src/plugins/dash/setup-text-tracks.js
+++ b/src/plugins/dash/setup-text-tracks.js
@@ -64,6 +64,10 @@ function attachDashTextTracksToVideojs(player, tech, tracks) {
 
     // Add track to videojs track list
     .map(({ trackConfig, dashTrack }) => {
+      if (dashTrack.isTTML && !player.getChild('TTMLTextTrackDisplay')) {
+        return null;
+      }
+
       const remoteTextTrack = player.addRemoteTextTrack(trackConfig, false);
 
       trackDictionary.push({ textTrack: remoteTextTrack.track, dashTrack });
@@ -76,6 +80,7 @@ function attachDashTextTracksToVideojs(player, tech, tracks) {
 
       return remoteTextTrack;
     })
+    .filter(el => el !== null)
 
   ;
 
--- a/src/plugins/dash/videojs-dash.js
+++ b/src/plugins/dash/videojs-dash.js
@@ -87,6 +87,24 @@ class Html5DashJS {
     // - mediaKeyMessageError (only fires under 'might not work' circumstances)
     // eslint-disable-next-line complexity
     this.retriggerError_ = (event) => {
+
+      if (event.type === 'error') {
+        // Initialize errors for ref https://cdn.dashjs.org/latest/jsdoc/core_errors_Errors.js.html
+        // Map to general init error
+        if (event.error.code >= 10 && event.error.code <= 35) {
+          this.player.error({ code: 4, dashjsErrorCode: event.error.code, message: event.error.message });
+        } else if (event.error.code < 10) { 
+          // Network errors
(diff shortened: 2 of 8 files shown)


src/plugins/dash/setup-text-tracks.js
```diff
@@ -64,6 +64,10 @@ function attachDashTextTracksToVideojs(player, tech, tracks) {
 
     // Add track to videojs track list
     .map(({ trackConfig, dashTrack }) => {
```
e1adn, line 66: Please consider replace map with reduce, so you don not need use filer and loop on array twice

src/plugins/dash/videojs-dash.js
```diff
@@ -94,20 +87,92 @@ class Html5DashJS {
...
-          this.player.error({ code: 5, dashjsErrorCode: event.error.code, message: event.error.message });
+      if (event.error === 'capability' && event.event === 'mediasource') {
+        // No support for MSE
+        this.player.error({
+          code: 4,
+          message: 'The media cannot be played because it requires a feature ' +
+            'that your browser does not support.'
+        });
+
+      } else if (event.error === 'manifestError' && (
```
e1adn, line 98: Maybe we should handle a map object from event id to error code and event message. I think it would be much cleaner instead of all those if.

e1adn: Please open an on-call for the devop on the oncall Slack channel to update the node to v12

e1adn (approved):

src/plugins/dash/videojs-dash.js
```diff
@@ -94,20 +87,92 @@ class Html5DashJS {
...
-
-        /* initialize errros for ref https://cdn.dashjs.org/latest/jsdoc/core_errors_Errors.js.html
-          map to general init error
-        */
-        if (event.error.code >= 10 && event.error.code <= 35) {
-          this.player.error({ code: 4, dashjsErrorCode: event.error.code, message: event.error.message });
-        } else if (event.error.code < 10) { // network errors
-          this.player.error({ code: event.error.code, dashjsErrorCode: event.error.code, message: event.error.message });
-        } else if (event.error.code >= 100 && event.error.code <= 114) { // Protection Errors  https://cdn.dashjs.org/latest/jsdoc/streaming_protection_errors_ProtectionErrors.js.html
-          this.player.error({ code: 5, dashjsErrorCode: event.error.code, message: event.error.message });
```
yaniv-cld, line 107: @tsi 
This part consolidated the dash errors to more general videojs errors so I think we should keep it

src/plugins/dash/videojs-dash.js
```diff
@@ -174,25 +243,27 @@ class Html5DashJS {
...
 
     // Attach the source with any protection data
     this.mediaPlayer_.setProtectionData(this.keySystemOptions_);
     this.mediaPlayer_.attachSource(manifestSource);
+
     this.tech_.triggerReady();
-    // map videojs seek
-    this.player.on(this.tech_, 'seeking', () => {
-      this.mediaPlayer_.seek(this.tech_.currentTime());
-    });
```
yaniv-cld, line 195: @tsi 
Seeking works without this?

yaniv-cld (commented):
Added comments

src/plugins/dash/videojs-dash.js
tsi: @yaniv-cld pls have a look at merging this together, I don't want to break any error handling we had here.

tsi:
TestNow

tsi:
testNow

yaniv-cld (approved):

PR #332, opened 2022-01-16
Two video controls are shown after exiting fullscreen mode

e1adn:
https://cloudinary.atlassian.net/browse/APPS-11584


e1adn:
@tsi you can check this on browserstack : 
https://jsfiddle.net/cld_elad/s17m6oq4/80/

PR #341, opened 2022-03-03
ads bug fix

e1adn:
- pick config param
- remove chainTarget prop from transformation object

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,14 +1,16 @@
 import 'assets/styles/main.scss';
 import VideoPlayer from './video-player';
 import { assign } from 'utils/assign';
-import { pick } from './utils/object';
+import { omit, pick } from './utils/object';
 import { CLOUDINARY_CONFIG_PARAM } from './video-player.const';
 
 if (window.cloudinary && window.cloudinary.Cloudinary) {
   console.warn('For version 1.9.0 and above, cloudinary-core is not needed for using the Cloudinary Video Player');
 }
 
-const getConfig = (playerOptions = {}, cloudinaryConfig) => assign(playerOptions, { cloudinaryConfig: cloudinaryConfig || pick(playerOptions, CLOUDINARY_CONFIG_PARAM) });
+const getConfig = (playerOptions = {}, cloudinaryConfig) => assign(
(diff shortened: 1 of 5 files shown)


src/plugins/cloudinary/common.js
```diff
@@ -50,7 +51,7 @@ const isSrcEqual = (source1, source2) => {
 
 export const extendCloudinaryConfig = (currentConfig, newConfig) => Object.assign(currentConfig, newConfig);
 
-export const getCloudinaryUrl = (publicId, transformation) => createCloudinaryLegacyURL(publicId, transformation);
+export const getCloudinaryUrl = (publicId, transformation) => createCloudinaryLegacyURL(publicId, omit(transformation, ['chainTarget']));
```
tsi, line 54: Let's figure out what's `chainTarget` and if it's even needed

src/index.js
```diff
@@ -1,12 +1,14 @@
...
 import { assign } from 'utils/assign';
+import { pick } from './utils/object';
+import { CLOUDINARY_CONFIG_PARAM } from './video-player.const';
 
 if (window.cloudinary && window.cloudinary.Cloudinary) {
   console.warn('For version 1.9.0 and above, cloudinary-core is not needed for using the Cloudinary Video Player');
 }
 
-const getConfig = (playerOptions = {}, cloudinaryConfig) => assign(playerOptions, { cloudinaryConfig: cloudinaryConfig || playerOptions });
+const getConfig = (playerOptions = {}, cloudinaryConfig) => assign(playerOptions, { cloudinaryConfig: cloudinaryConfig || pick(playerOptions, CLOUDINARY_CONFIG_PARAM) });
```
tsi, line 11: this is becoming quite confusing.
if we call `getConfig({cloud_name: 'foo'})`
you'll get in return:
```
{
  cloud_name: 'foo'
  cloudinaryConfig: {
    cloud_name: 'foo'
  }
}
```
correct? is that what we want?

src/plugins/cloudinary/common.js
e1adn: its need for chaining , so you can do:
player.transformation().source

tsi (approved):

PR #352, opened 2022-08-04
POP-1125-playlist-next

tsi:
This PR prevents a situation where the button click propagate to the clickable wrapper element and that caused the click handler to be called twice
Fixes https://cloudinary.atlassian.net/browse/POP-1125


e1adn:
💪

e1adn (approved):

PR #364, opened 2022-11-28
POP-8186-transformation-arrays-case

tsi:
Jira: https://cloudinary.atlassian.net/browse/ME-4322

The Transformation class is transforming keys inside transformation objects, but not in a transformation array
See https://codesandbox.io/s/js-sdk-transformation-legacyurl-b64xur?file=/src/index.js

This PR makes sure arrays are parsed separately before being sent


ehab-cl:
@tsi is this still relevant so we can merge? or we should close it?

tsi:
Not sure what to say, this is a partial solution for a bigger problem, see the [discussion](https://cloudinary.slack.com/archives/CRGL38TTJ/p1668431345261859) linked in the ticket

PR #374, opened 2023-01-09
ME-4804-prevent-error-while-lazy-loading

tsi:
Story: https://cloudinary.atlassian.net/browse/ME-4802

### DIFF
--- a/src/components/interaction-area/interaction-area.service.js
+++ b/src/components/interaction-area/interaction-area.service.js
@@ -15,7 +15,7 @@ import {
   updateInteractionAreasItem
 } from './interaction-area.utils';
 import { addEventListener, createElement } from '../../utils/dom';
-import { throttle } from '../../utils/time';
+import { throttle } from '../../utils/throttle';
 import { get } from '../../utils/object';
 import { noop } from '../../utils/type-inference';
 import { addMetadataTrack } from '../../video-player.utils';
--- a/src/video-player.js
+++ b/src/video-player.js
@@ -11,7 +11,6 @@ import qualitySelector from './components/qualitySelector/qualitySelector.js';
 // #endif
 import VideoSource from './plugins/cloudinary/models/video-source/video-source';
 import { isFunction, isString, isPlainObject } from './utils/type-inference';
-import { throttle } from './utils/throttle';
 import {
   extractOptions,
   getResolveVideoElement,
@@ -215,9 +214,7 @@ class VideoPlayer extends Utils.mixin(Eventable) {
 
     await this._initDash(options);
 
-    throttle(() => {
-      this.loadingLazyPlugins = false;
-    });
+    this.loadingLazyPlugins = false;
   }
 
   async _initDash(options) {
--- a/src/video-player.js
+++ b/src/video-player.js
@@ -11,6 +11,7 @@ import qualitySelector from './components/qualitySelector/qualitySelector.js';
 // #endif
 import VideoSource from './plugins/cloudinary/models/video-source/video-source';
 import { isFunction, isString, isPlainObject } from './utils/type-inference';
+import { throttle } from './utils/throttle';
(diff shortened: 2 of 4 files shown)


src/video-player.js
```diff
@@ -201,7 +211,13 @@ class VideoPlayer extends Utils.mixin(Eventable) {
   }
 
   async _initLazyPlugins(options) {
+    this.loadingLazyPlugins = true;
+
     await this._initDash(options);
+
+    throttle(() => {
+      this.loadingLazyPlugins = false;
+    });
```
ehab-cl, line 220: I have 3 questions here:
- I believe we should be sending a timeout argument here, since the throttle function has no default timeout, right?
- I think this will never be executed eventually this will just create a function that does not ever get executed 
- I am curious why do you want to throttle this? it seems to me throttling it will actually be more expensive both memory and cpu over just executing it directly

ehab-cl (approved):
💪

src/components/interaction-area/interaction-area.service.js
```diff
@@ -15,7 +15,7 @@ import {
   updateInteractionAreasItem
 } from './interaction-area.utils';
 import { addEventListener, createElement } from '../../utils/dom';
-import { throttle } from '../../utils/time';
+import { throttle } from '../../utils/throttle';
```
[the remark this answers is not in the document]
tsi: I agree in general, but I already had a `throttle()`, actually, I had two :)

src/video-player.js
```diff
@@ -125,6 +126,15 @@ class VideoPlayer extends Utils.mixin(Eventable) {
 
   _setVideoJsListeners(ready) {
 
+    // Prevent flash of error message while lazy-loading plugins.
+    videojs.hook('beforeerror', (player, err) => {
+      if (err && err.code === 3 && this.loadingLazyPlugins) {
```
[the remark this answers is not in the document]
tsi: `code == 3` means a media issue i.e. corrupted manifest etc
I used it like this because I saw it being used this way in a few other places, I'm in for refactoring this, but I don't think this is the place to do that.

src/video-player.js
ehab-cl: @tsi i might be completely wrong, but just based on code it is hard for me to understand how your function gets called, like is throttle calling the input function immediately? 

Anyway if you ask me based on your description it seems you are using throttle just as a way to setTimeout, both are indirect workarounds in my eyes, but i believe setTimeout is more explicit.

tsi: @ehab-cl you are correct, both about the settimeout and about it being a workaround. I didn't like it but I needed a way to wait for the cycle to end for the error to never appear.
Anyway, pls see my latest version, IMO it makes much more sense.

src/video-player.js
```diff
@@ -215,9 +214,7 @@ class VideoPlayer extends Utils.mixin(Eventable) {
 
     await this._initDash(options);
 
-    throttle(() => {
-      this.loadingLazyPlugins = false;
-    });
+    this.loadingLazyPlugins = false;
```
ehab-cl, line 217: I think enclosing this in a setTimeout is more explicit than what you did at the dashPlugin function - essentially per my understanding the issue is here, you want this to kinda of be executed in the next cycle, i think having the workaround closer to the issue makes it easier to understand and harder to just lose by mistake when refactoring/changing.

ehab-cl (approved):
Added one minor suggestion

src/video-player.js
tsi: It might be more clear, yes. but the effect here is not exactly the same.
This time I'm postponing the resolved state of the `dashPlugin` promise, instead of just postponing setting the `this.loadingLazyPlugins` value while keep running (which is what I did in the previous version).
To me this looks less prone to race-condition issues.

PR #387, opened 2023-03-14
fix: set aria hidden for some empty elements

jakub-roch:
Set `aria-hidden=true` for empty headers & p elements.

### DIFF
--- a/src/components/recommendations-overlay/recommendations-overlay-primary-item.js
+++ b/src/components/recommendations-overlay/recommendations-overlay-primary-item.js
@@ -55,12 +55,15 @@ class RecommendationsOverlayPrimaryItem extends RecommendationsOverlayItem {
     });
 
     this.title = dom.createEl('h2');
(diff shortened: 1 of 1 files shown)


jakub-roch:
> This is great! but shouldn't we change this when the elements do appear?

anyway if anything is added to HTML its being checked regarding accessibility so Im not really sure if we need to limit it only to visible elements

ehab-cl:
I might be wrong, but when we have that attribute on non empty elements i think this will be hurting accessibility especially for header elements, wdyt Kuba (especially if this element provides actual value)? 

(Also if feasible it is probably better to not render the empty element at all this way it wont appear in the accessibility tree and in the DOM as well)

jakub-roch:
actually you are right guys, thank you! I changed it to check if there is some text inside - if not we are adding aria-hidden element (initially it's being set cuz it's empty) @ehab-cl @tsi

tsi (approved):
Looking good 👍

PR #393, opened 2023-06-11
ME-5927-source-syntax

tsi:


src/video-player.js
```diff
@@ -545,6 +545,8 @@ class VideoPlayer extends Utils.mixin(Eventable) {
   }
 
   source(publicId, options = {}) {
+    ({ publicId, options } = normalizeOptions(publicId, options));
```
ehab-cl, line 548: That function was there already, if i have to guess it was maybe used here or somewhere else and someone removed it either by mistake or intentionally, if intentionally do we know why?

Like just trying to bring awareness to the possible idea that someone might have removed this intentionally in-order to fix something.

ehab-cl (approved):
💪

PR #399, opened 2023-07-03
ME-6565-seek-thumbs-by-default-2

tsi:
This PR fixes seek-thumbs compatibility with playlist coming-up & recommendations

### DIFF
--- a/src/plugins/videojs-http-source-selector/plugin.js
+++ b/src/plugins/videojs-http-source-selector/plugin.js
@@ -46,9 +46,13 @@ const onPlayerReady = (player) => {
     } else {
       console.log('player.videojs_http_source_selector_initialized == false');
       player.videojs_http_source_selector_initialized = true;
-      const controlBar = player.controlBar,
-        fullscreenToggle = controlBar.getChild('fullscreenToggle').el();
-      controlBar.el().insertBefore(controlBar.addChild('SourceMenuButton').el(), fullscreenToggle);
+      const controlBar = player.controlBar;
+      const fullscreenToggle = controlBar.getChild('fullscreenToggle');
+      if (fullscreenToggle) {
+        controlBar.el().insertBefore(controlBar.addChild('SourceMenuButton').el(), fullscreenToggle.el());
+      } else {
+        controlBar.el().appendChild(controlBar.addChild('SourceMenuButton').el());
+      }
     }
   });
 };
--- a/src/plugins/vtt-thumbnails/vtt-thumbnails.scss
+++ b/src/plugins/vtt-thumbnails/vtt-thumbnails.scss
@@ -11,6 +11,7 @@
   // border: 3px solid var(--color-base);
   border-radius: 3px;
   box-shadow: 0 0 3px rgba(0, 0, 0, 0.3);
+  backdrop-filter: blur(12px);
 }
 
 .vjs-vtt-time-display {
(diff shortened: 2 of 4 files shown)


src/plugins/vtt-thumbnails/vtt-thumbnails.scss
```diff
@@ -11,6 +11,7 @@
   // border: 3px solid var(--color-base);
   border-radius: 3px;
   box-shadow: 0 0 3px rgba(0, 0, 0, 0.3);
+  backdrop-filter: blur(12px);
```
tsi, line 14: While the browser fetch the thumbnails, blur what's behind it so it looks better

src/video-player.js
```diff
@@ -306,16 +306,16 @@ class VideoPlayer extends Utils.mixin(Eventable) {
 
         const publicId = source.publicId();
 
-        const transformations = source.transformation();
```
tsi, line 309: This fixes the compatibility issue

src/video-player.js
```diff
@@ -306,16 +306,16 @@ class VideoPlayer extends Utils.mixin(Eventable) {
...
+        const transformation = Utils.assign({}, source.transformation());
```
tsi, line 309: This fixes the compatibility issue

src/plugins/videojs-http-source-selector/plugin.js
```diff
@@ -46,9 +46,13 @@ const onPlayerReady = (player) => {
...
-      const controlBar = player.controlBar,
-        fullscreenToggle = controlBar.getChild('fullscreenToggle').el();
-      controlBar.el().insertBefore(controlBar.addChild('SourceMenuButton').el(), fullscreenToggle);
+      const controlBar = player.controlBar;
+      const fullscreenToggle = controlBar.getChild('fullscreenToggle');
+      if (fullscreenToggle) {
+        controlBar.el().insertBefore(controlBar.addChild('SourceMenuButton').el(), fullscreenToggle.el());
+      } else {
+        controlBar.el().appendChild(controlBar.addChild('SourceMenuButton').el());
+      }
```
tsi, line 55: This also triggered an error in our custom components page since `seekThumbnails` assumes there's a progress bar (which makes sense), and another one since `http-source-selector` assumes there's a `fullscreenToggle`

ehab-cl: ok but i do not see you are checking here whether seekThumbnails is enabled or not?

src/video-player.js
ehab-cl: what is the compatibility issue you are referring to here?

ehab-cl (approved):
💪

src/video-player.js
tsi: The original issue that re-opened this - when seekThumbs is enabled it breaks playlist "coming-up" component & recommendations overlay.

src/plugins/videojs-http-source-selector/plugin.js
tsi: yeah so we had two issues on that page - the first one was because this example removes the progress-bar and `seekThumbnails` assumes it exists, fixed by disabling seekThumbs in this example. The second one then showed up because the quality-selector assumes the full-screen toggle exists, which is also wrong (but not directly related to seek-thumbnails)

PR #403, opened 2023-07-16
ME-6552-lazy-ads - lazy load 'videojs-contrib-ads'

tsi:
This PR adds `videojs-contrib-ads` to the lazy loaded module, instead of bundling it to the main one


ehab-cl (approved):
💪

PR #404, opened 2023-07-17
Me 6690 prevent undefined cloud name

jakub-roch:
Prevent sending undefined or invalid values from client side (simple, temporary solution - before we connect VP with new cld analytics package).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+1.9.11-edge.2 / 2023-07-03
+==================
+
+* Fix: seek-thumbnail compatibility with other plugins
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,20 +1,9 @@
...
-1.9.11-edge.1 / 2023-06-26
-==================
-
-Changes
--------------
-- Turn seek-thumbnail on by default
-
-1.9.11-edge.0 / 2023-06-18
-==================
-
```
tsi, line 17: @jakub-roch this shouldn't happen. could it be that you started from `master` instead of `edge` ?

PR #408, opened 2023-07-25
fix: prevent sending undefined cloud name from cld analytics

jakub-roch:
Prevent sending undefined as cloud name in cld analytics


tsi (approved):
👍

PR #410, opened 2023-08-01
ME-6929-histogram-plugin

tsi:
This PR adds a plugin to display a histogram chart on the progress bar.

### DIFF
--- a/src/plugins/histogram/index.js
+++ b/src/plugins/histogram/index.js
@@ -161,7 +161,6 @@ const HistogramPlugin = (function () {
     // Create a smooth line path
     const path = document.createElementNS('http://www.w3.org/2000/svg', 'path');
     path.setAttribute('fill', 'lightblue');
-    // path.setAttribute('stroke', 'blue');
 
     // Generate the smooth line path data
     let d = `M ${points[0].x},${points[0].y}`;
@@ -190,12 +189,9 @@ const HistogramPlugin = (function () {
     svg.setAttribute('preserveAspectRatio', 'none');
     svg.setAttribute('width', svgWidth);
     svg.setAttribute('height', svgHeight);
+    svg.innerHTML = '';
 
     const path = this.createPath(data, svgWidth, svgHeight);
-
-    while (svg.firstChild) {
-      svg.removeChild(svg.firstChild);
-    }
     svg.appendChild(path);
 
     this.histogramHolder.appendChild(svg);
--- a/src/plugins/histogram/index.js
+++ b/src/plugins/histogram/index.js
@@ -61,7 +61,6 @@ const HistogramPlugin = (function () {
   function HistogramPlugin(player, options) {
     this.player = player;
     this.options = options;
-    this.listenForDurationChange();
     this.initializeHistogram();
     this.registeredEvents = {};
     return this;
@@ -86,16 +85,9 @@ const HistogramPlugin = (function () {
     delete this.lastStyle;
   };
 
-  HistogramPlugin.prototype.listenForDurationChange = function listenForDurationChange() {
-    this.player.on('durationchange', function () {
-      // ToDo
-    });
-  };
-
   /**
    * Bootstrap the plugin.
    */
-
   HistogramPlugin.prototype.initializeHistogram = function initializeHistogram() {
     const _this2 = this;
 
--- a/src/plugins/histogram/index.js
+++ b/src/plugins/histogram/index.js
@@ -33,7 +33,7 @@ const onPlayerReady = function onPlayerReady(player, options) {
  *           An object of options left to the plugin author to define.
  */
 const histogram = function histogram(options) {
-  let _this = this;
+  const _this = this;
 
   this.ready(function () {
     onPlayerReady(_this, videojs.mergeOptions(defaults, options));
@@ -97,7 +97,7 @@ const HistogramPlugin = (function () {
    */
 
   HistogramPlugin.prototype.initializeHistogram = function initializeHistogram() {
-    let _this2 = this;
+    const _this2 = this;
 
     if (!this.options.src) {
       return;
@@ -119,10 +119,10 @@ const HistogramPlugin = (function () {
    */
 
   HistogramPlugin.prototype.getHistogramFile = function getHistogramFile(url) {
-    let _this3 = this;
+    const _this3 = this;
 
     return new Promise(function (resolve) {
-      let req = new XMLHttpRequest();
+      const req = new XMLHttpRequest();
       req.data = {
         resolve: resolve
       };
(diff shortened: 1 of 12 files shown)


src/plugins/histogram/index.js
```diff
@@ -0,0 +1,215 @@
+import videojs from 'video.js';
+
+import './histogram.scss';
+
+// Default options for the plugin.
+let defaults = {};
```
ehab-cl, line 6:
```suggestion
const defaults = {};
```

src/plugins/histogram/index.js
```diff
@@ -0,0 +1,215 @@
...
+      this.histogramHolder.parentNode.removeChild(this.histogramHolder);
+    }
+    delete this.progressBar;
+    delete this.histogramHolder;
+    delete this.lastStyle;
+  };
+
+  HistogramPlugin.prototype.listenForDurationChange = function listenForDurationChange() {
+    this.player.on('durationchange', function () {
+      // ToDo
```
ehab-cl, line 91: ?

src/plugins/histogram/index.js
```diff
@@ -0,0 +1,215 @@
...
+    onPlayerReady(_this, videojs.mergeOptions(defaults, options));
+  });
+};
+
+/**
+ * Histogram class.
+ *
+ * This class performs all functions related to displaying the Histogram bar.
+ */
+const HistogramPlugin = (function () {
```
ehab-cl, line 48: For next time i recommend writing this using Es6 class syntax

src/plugins/histogram/index.js
```diff
@@ -0,0 +1,215 @@
...
+    if (!this.options.src) {
+      return;
+    }
+
+    this.getHistogramFile(this.options.src).then(function (res) {
+      _this2.setupHistogramElement();
+      if (_this2.histogramHolder) {
+        _this2.createHistogram(res);
+      }
+    });
```
ehab-cl, line 111: You need this2 because of the dynamic binding of `this` keyword  that happens in a regular function, if you use an arrow function here you can get rid of the confusing `this2` declaration and just use regular `this` that will be lexically bound, which i think will work.
```suggestion
    this.getHistogramFile(this.options.src).then(res => {
      this.setupHistogramElement();
      if (this.histogramHolder) {
        this.createHistogram(res);
      }
    });
```

src/plugins/histogram/index.js
```diff
@@ -0,0 +1,215 @@
...
+
+  /**
+   * Grabs the contents of the VTT file.
+   *
+   * @param url
+   * @returns {Promise}
+   */
+
+  HistogramPlugin.prototype.getHistogramFile = function getHistogramFile(url) {
+    let _this3 = this;
```
ehab-cl, line 122: see my comment above on how to get rid of this declaration.

src/plugins/histogram/index.js
```diff
@@ -0,0 +1,215 @@
...
+
+    return new Promise(function (resolve) {
```
ehab-cl, line 124: why not just use fetch and async/await?

src/plugins/histogram/index.js
```diff
@@ -0,0 +1,215 @@
...
+  /**
+   * Callback for loaded Histogram file.
+   */
+
+  HistogramPlugin.prototype.histogramDataLoaded = function histogramDataLoaded() {
+    this.data.resolve(JSON.parse(this.responseText));
+  };
+
+  HistogramPlugin.prototype.setupHistogramElement = function setupHistogramElement() {
+    let mouseDisplay = this.player.$('.vjs-mouse-display');
```
ehab-cl, line 144: Please replace let declarations with constant declarations where possible

src/plugins/histogram/index.js
```diff
@@ -0,0 +1,215 @@
...
+    const stepX = containerWidth / (dataArray.length - 1);
+    const points = dataArray.map((value, index) => ({
+      x: index * stepX,
+      y: containerHeight - value * containerHeight
+    }));
+
+    // Create a smooth line path
+    const path = document.createElementNS('http://www.w3.org/2000/svg', 'path');
+    path.setAttribute('fill', 'lightblue');
+    // path.setAttribute('stroke', 'blue');
```
ehab-cl, line 172:
```suggestion
```

src/plugins/histogram/index.js
```diff
@@ -0,0 +1,215 @@
...
+
+    path.setAttribute('d', d);
+
+    return path;
+  };
+
+  HistogramPlugin.prototype.createHistogram = function createHistogram(info) {
+    const data = info.data;
+    const svgWidth = 600;
+    const svgHeight = 20;
```
ehab-cl, line 194: should not these be dynamic based on the video element size?

src/plugins/histogram/index.js
```diff
@@ -0,0 +1,215 @@
...
+    svg.setAttribute('viewBox', `0 0 ${svgWidth} ${svgHeight}`);
+    svg.setAttribute('preserveAspectRatio', 'none');
+    svg.setAttribute('width', svgWidth);
+    svg.setAttribute('height', svgHeight);
+
+    const path = this.createPath(data, svgWidth, svgHeight);
+
+    while (svg.firstChild) {
+      svg.removeChild(svg.firstChild);
+    }
```
ehab-cl, line 206: It seems to me you are just emptying the inner html, this is more safe than a while loop, and it is much more faster since it will cause only one layout shift calculation instead of possibly few ones. 

```suggestion
   svg.innerHTML = ''
```

src/plugins/vtt-thumbnails/index.js
```diff
@@ -190,7 +190,8 @@ const VttThumbnailsPlugin = (function () {
 
     let mouseDisplay = this.player.$('.vjs-mouse-display');
     this.progressBar = this.player.$('.vjs-progress-control');
-    let thumbHolder = document.createElement('div');
+    let thumbHolder = this.player.$('.vjs-vtt-thumbnail-display') || document.createElement('div');
```
ehab-cl, line 193:
```suggestion
    const thumbHolder = this.player.$('.vjs-vtt-thumbnail-display') || document.createElement('div');
```

ehab-cl (approved):
💪 LGTM

src/plugins/histogram/index.js
tsi: This is building on top of the seek-thumbnails "boilerplate", so uses the same structure, we can refactor, but this was quite helpful and easy to use

src/plugins/histogram/index.js
tsi: no, this is setting the SVG aspect ratio, rendered size is stretched to 100% width

ehab-cl (approved):
LGTM 💪

PR #416, opened 2023-08-23
ME-4810-support-cloudName

tsi:
This PR adds support for camelCase cld configuration alongside snake_case configs like it did before.

https://cloudinary.atlassian.net/browse/ME-4810

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,22 +1,29 @@
 import 'assets/styles/main.scss';
 import VideoPlayer from './video-player';
 import { assign } from 'utils/assign';
-import { omit, pick } from './utils/object';
+import { omit, pick, convertKeysToSnakeCase } from './utils/object';
 import { CLOUDINARY_CONFIG_PARAM } from './video-player.const';
 
 if (window.cloudinary && window.cloudinary.Cloudinary) {
-  console.warn('For version 1.9.0 and above, cloudinary-core is not needed for using the Cloudinary Video Player');
+  console.warn(
+    'For version 1.9.0 and above, cloudinary-core is not needed for using the Cloudinary Video Player'
+  );
 }
 
-const getConfig = (playerOptions = {}, cloudinaryConfig) => assign(
-  omit(playerOptions, CLOUDINARY_CONFIG_PARAM),
-  { cloudinaryConfig: cloudinaryConfig || pick(playerOptions, CLOUDINARY_CONFIG_PARAM) });
+const getConfig = (playerOptions = {}, cloudinaryConfig) =>
+  // pick cld-configurations and assign them to cloudinaryConfig
+  assign(omit(playerOptions, CLOUDINARY_CONFIG_PARAM), {
+    cloudinaryConfig:
+      cloudinaryConfig || convertKeysToSnakeCase(pick(playerOptions, CLOUDINARY_CONFIG_PARAM))
+  });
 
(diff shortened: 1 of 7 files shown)


src/utils/object.js
```diff
@@ -49,3 +49,14 @@ export const omit = (obj, keys) => {
...
   }, {});
 };
+
+export const convertKeysToSnakeCase = (obj) => {
+  let snakeCaseObj = {};
+
+  Object.keys(obj).forEach((key) => {
+    const snakeCaseKey = camelToSnake(key);
+    snakeCaseObj[snakeCaseKey] = obj[key];
+  });
```
ehab-cl, line 59:
```suggestion
for (const key of Object.keys(obj)) {
    const snakeCaseKey = camelToSnake(key);
    snakeCaseObj[snakeCaseKey] = obj[key];
}
```

src/utils/string.js
```diff
@@ -12,4 +11,10 @@ function startsWith(str, searchStr, position) {
   }
 }
 
-export { camelize, startsWith };
+function camelToSnake(camelCaseString) {
```
ehab-cl, line 14: I think we should use Lodash functions, with tree shaking importing single functions is minimal on the payload size, on the other hand it is a huge boost for safety and readability, and sometimes even for performance.

I would also use arrow functions here for declaring these functions. If you are also implementing it yourself i would introduce unit tests, btw do we have any unit tests for the VP?

src/video-player.const.js
```diff
@@ -50,7 +51,8 @@ export const CLOUDINARY_CONFIG_PARAM = [
   'url_suffix',
   'use_root_path',
   'auth_token'
-];
+  // Support camelCase as well
+].map((param) => ([param, camelize(param)])).flat();
```
ehab-cl, line 55: see my first comment since it can remove the need for doing this, but anyway i personally would not do it programmatically but would just introduce it manually, more code but easier to read IMO and more safe.

ehab-cl (approved):
💪

src/index.js
```diff
@@ -1,22 +1,29 @@
...
 }
 
-const getConfig = (playerOptions = {}, cloudinaryConfig) => assign(
-  omit(playerOptions, CLOUDINARY_CONFIG_PARAM),
-  { cloudinaryConfig: cloudinaryConfig || pick(playerOptions, CLOUDINARY_CONFIG_PARAM) });
+const getConfig = (playerOptions = {}, cloudinaryConfig) =>
+  // pick cld-configurations and assign them to cloudinaryConfig
+  assign(omit(playerOptions, CLOUDINARY_CONFIG_PARAM), {
+    cloudinaryConfig:
+      cloudinaryConfig || convertKeysToSnakeCase(pick(playerOptions, CLOUDINARY_CONFIG_PARAM))
```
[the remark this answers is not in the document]
tsi: I actually like that, but as it is, the `omit` will not work as intended. this will cause the snakeCase `cloudName` to be passed to the player `source()` method. but I'm looking into that still...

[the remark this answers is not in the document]
tsi: looking on this again I'm not sure we even need to `omit` anything, since this goes through `extractOptions()` in `video-player`'s constructor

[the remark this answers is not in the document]
tsi: Done ✅

src/utils/string.js
tsi: Added Lodash

PR #417, opened 2023-08-27
feat: integrate new cld analytics package

jakub-roch:
Integrate new package of cld-analytics with VP.


package.json
```diff
@@ -65,6 +65,7 @@
   },
   "dependencies": {
     "@cloudinary/url-gen": "^1.10.2",
+    "cloudinary-video-analytics": "1.1.0",
```
ehab-cl, line 68: We should aim the keep semantic versioning there, and so it is a good idea to get minor versions here. 
```suggestion
    "cloudinary-video-analytics": "^1.1.0",
```

src/plugins/cloudinary-analytics/index.js
```diff
@@ -1,19 +1,24 @@
-import { trackVideoPlayer } from '../../components/cld-analytics/cld-analytics';
+import connectCloudinaryAnalytics from 'cloudinary-video-analytics';
```
ehab-cl, line 1: i think it is better to make our package export a named entity, wdyt? 
i mean defaults are problematic, tomorrow if you decide to export another thing it must be named, and so this forces a mixed syntax for the default and named when both exporting and importing. Also this will make naming more consistent.

ehab-cl (approved):
LGTM 💪

package.json
jakub-roch: 👍

src/plugins/cloudinary-analytics/index.js
jakub-roch: I agree but I it should be done in another task - https://cloudinary.atlassian.net/browse/ME-13442

PR #426, opened 2023-09-05
fix: add missing props to init request

jakub-roch:
* add missing `cloudName` to init request
* make init request ONLY when there is source (playlist/source)
* change param `vp_version` to `vpVersion`

### DIFF
--- a/src/video-player.js
+++ b/src/video-player.js
@@ -118,16 +118,26 @@ class VideoPlayer extends Utils.mixin(Eventable) {
     this._initPlaylistWidget();
     this._initJumpButtons();
     this._setVideoJsListeners(ready);
-    this._sendAnalytics(this.playerOptions);
   }
 
-  _sendAnalytics(options) {
+  _sendInternalAnalytics(additionalOptions = {}) {
+    if (this._internalAnalyticsSent) {
+      return;
+    }
+
     try {
+      const cloudName = this.cloudinaryConfig().cloud_name;
+      const options = Utils.assign({}, this.playerOptions, additionalOptions);
       const analyticsData = getAnalyticsFromPlayerOptions(options);
-      const qs = new URLSearchParams(analyticsData).toString();
-      fetch(`${INTERNAL_ANALYTICS_URL}/video_player_init?${qs}&vp_version=${VERSION}`);
+      const analyticsParams = new URLSearchParams(analyticsData).toString();
+      const baseParams = new URLSearchParams({
+        vpVersion: VERSION,
+        cloudName
+      }).toString();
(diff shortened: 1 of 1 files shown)


src/video-player.js
```diff
@@ -118,16 +118,26 @@ class VideoPlayer extends Utils.mixin(Eventable) {
...
     this._initJumpButtons();
     this._setVideoJsListeners(ready);
-    this._sendAnalytics(this.playerOptions);
   }
 
-  _sendAnalytics(options) {
+  _sendInternalAnalytics(additionalOptions = {}) {
+    if (this._internalAnalyticsSent) {
+      return;
+    }
```
ehab-cl, line 126: It seems to me that if we remove this if statement we will get the usage collected each time the source is changed right? if so, and only that is required to make it work, with no risks, maybe we should supporting the analytics whenever the source is changing?

src/video-player.js
```diff
@@ -612,6 +624,7 @@ class VideoPlayer extends Utils.mixin(Eventable) {
 
   playlist(sources, options = {}) {
     this._initQualitySelector();
+    this._sendInternalAnalytics();
```
ehab-cl, line 627: should not we send any param here?

ehab-cl (approved):
LGTM 💪

src/video-player.js
```diff
@@ -118,16 +118,26 @@ class VideoPlayer extends Utils.mixin(Eventable) {
...
   }
 
-  _sendAnalytics(options) {
+  _sendInternalAnalytics(additionalOptions = {}) {
+    if (this._internalAnalyticsSent) {
+      return;
+    }
+
     try {
+      const cloudName = this.cloudinaryConfig().cloud_name;
```
tsi, line 129: Why not do it together with everything else inside `getCloudinaryOptions`? this seems to unnecessarily complicate this.

src/video-player.js
jakub-roch: we get there options that are not listed in analytics params for init action

src/video-player.js
jakub-roch: actually thats question for Raz - init event should be send only once, otherwise it has totally different meaning, maybe we should rename it to `video_player_source` ?

src/video-player.js
jakub-roch: do you mean to put baseParams into `getAnalyticsFromPlayerOptions` ? If so it would mean we are expanding responsibility of this function to have access to video player instance to get those data - or we would simply pass it there which IMO is similar to what is there

jakub-roch: I changed it to get cloud name directly `cloudName: options.cloudinary.cloudinaryConfig.cloud_name` but we need to keep it separated because of internal fn `getInstanceId`

PR #437, opened 2023-10-04
ME-6707-chapters

tsi:
This PR introduces a new chapters plugin.

ToDo:
* ✅ Refactor vtt-thumbnails to use core videojs functionality - mouse tracking, DOM components
* ✅ Progress bar markers
* ✅ Progress bar chapters display (on hover)
* ✅ Control bar current chapter display
* ✅ Allow inline chapters (VTT alternative)
* ✅ Look into console warnings

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,6 @@
     "video.js": "8.5.2",
     "videojs-contrib-ads": "^6.9.0",
     "videojs-contrib-dash": "^5.1.1",
-    "videojs-contrib-quality-levels": "^4.0.0",
     "videojs-ima": "^2.1.0",
     "videojs-per-source-behaviors": "^3.0.0",
     "webfontloader": "^1.6.28"
--- a/package.json
+++ b/package.json
@@ -73,7 +73,6 @@
     "videojs-contrib-ads": "^6.9.0",
     "videojs-contrib-dash": "^5.1.1",
     "videojs-ima": "^2.1.0",
-    "videojs-per-source-behaviors": "^3.0.0",
     "webfontloader": "^1.6.28"
   },
   "devDependencies": {
--- a/src/plugins/chapters/index.js
+++ b/src/plugins/chapters/index.js
@@ -84,6 +84,7 @@ const ChaptersPlugin = (function () {
    */
   ChaptersPlugin.prototype.initializeChapters = function initializeChapters() {
     if (this.options.url) {
+      // Fetch chapters VTT from URL
       const chaptersTrack = {
         kind: 'chapters',
         src: this.options.url
@@ -91,11 +92,10 @@ const ChaptersPlugin = (function () {
       const textTrack = this.player.addRemoteTextTrack(chaptersTrack);
       textTrack.addEventListener('load', () => {
         this.chaptersTrack = textTrack.track;
-        this.setupProgressBarMarkers();
-        this.setupProgressBarChapter();
-        this.setupControlBarChapter();
+        this.setupChaptersDisplays();
       });
     } else {
+      // Setup chapters from options
       const textTrack = this.player.addTextTrack('chapters');
       const end = this.player.duration();
       Object.entries(this.options).forEach((entry, index, arr) => {
@@ -107,10 +107,17 @@ const ChaptersPlugin = (function () {
         textTrack.addCue(cue);
       });
       this.chaptersTrack = textTrack;
-      this.setupProgressBarMarkers();
-      this.setupProgressBarChapter();
-      this.setupControlBarChapter();
+      this.setupChaptersDisplays();
     }
+  };
+
+  /**
+   * Setup chapter displays.
+   */
+  ChaptersPlugin.prototype.setupChaptersDisplays = function initializeChapters() {
+    this.setupProgressBarMarkers();
+    this.setupProgressBarChapter();
+    this.setupControlBarChapter();
   };
 
   /**
--- a/src/plugins/chapters/index.js
+++ b/src/plugins/chapters/index.js
@@ -162,7 +162,7 @@ const ChaptersPlugin = (function () {
         'timeTooltip'
       ]);
 
-      timeTooltip.el().parentElement.appendChild(chapterEl);
+      timeTooltip.el().parentElement.prepend(chapterEl);
 
       const getChapterFromPoint = point => {
         const total = this.player.duration();
--- a/src/plugins/chapters/index.js
+++ b/src/plugins/chapters/index.js
@@ -1,5 +1,3 @@
-// Todo: eslint
-/* eslint-disable */
 import videojs from 'video.js';
 
 import './chapters.scss';
@@ -48,6 +46,7 @@ const chapters = function chapters(options) {
  * Chapters
  */
 const ChaptersPlugin = (function () {
+
   /**
    * Plugin class constructor, called by videojs on
    * ready event.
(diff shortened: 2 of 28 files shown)


tsi:
Chapters can be configured in two ways:

1. with a VTT file:
```
const player = cloudinary.videoPlayer('player', {
  cloudName: 'demo',
  publicId: 'cld-imagine',
  chapters: {
    url: 'https://res.cloudinary.com/tsachi/raw/upload/tests/snow_horses_chapters_k1e1ff.vtt'
  }
});
```

2. inline chapters object:
```
const player = cloudinary.videoPlayer('player', {
  cloudName: 'demo',
  publicId: 'cld-imagine',
  chapters: {
    0: 'Chapter 1',
    50: 'Chapter 2',
    80: 'Chapter 3'
  }
});
```

package.json
```diff
@@ -72,7 +72,6 @@
     "video.js": "8.5.2",
     "videojs-contrib-ads": "^6.9.0",
     "videojs-contrib-dash": "^5.1.1",
-    "videojs-contrib-quality-levels": "^4.0.0",
```
tsi, line 75: This was throwing a warning, turns out it is included now in videojs and not needed from our end

ehab-cl:
> Chapters can be configured in two ways:
> 
> 1. with a VTT file:
> 
> ```
> const player = cloudinary.videoPlayer('player', {
>   cloudName: 'demo',
>   publicId: 'cld-imagine',
>   chapters: {
>     url: 'https://res.cloudinary.com/tsachi/raw/upload/tests/snow_horses_chapters_k1e1ff.vtt'
>   }
> });
> ```
> 
> 2. inline chapters object:
> 
> ```
> const player = cloudinary.videoPlayer('player', {
>   cloudName: 'demo',
>   publicId: 'cld-imagine',
>   chapters: {
>     0: 'Chapter 1',
>     50: 'Chapter 2',
>     80: 'Chapter 3'
>   }
> });
> ```

what does the object keys represent, seconds? or milliseconds? 

I have also thought more about this, are you sure we want to create a non standard way to support chapters, supporting and maintaining it? i genuinely think that for the long term it will be better to stick only with vtt...it is a standard way and should be enough - but it is your call eventually.

src/plugins/chapters/index.js
```diff
@@ -0,0 +1,195 @@
...
+      'timeTooltip'
+    ]);
+
+    timeTooltip.el().parentElement.prepend(chapterEl);
+
+    const getChapterFromPoint = point => {
+      const total = this.player.duration();
+      const seekBarTime = point * total;
+      const chapter = this.chaptersTrack.cues_.find(marker => {
+        return marker.startTime <= seekBarTime && marker.endTime >= seekBarTime;
```
ehab-cl, line 180: Very subjective, but if you ask me this becomes much easier to read, it becomes clear that you are making sure that seekBarTime is between startTime and endTime 
```suggestion
        return seekBarTime >= marker.startTime  &&  seekBarTime  <= marker.endTime;
```

src/plugins/chapters/index.js
```diff
@@ -0,0 +1,195 @@
...
+      });
+      return chapter ? chapter.text : '';
```
ehab-cl, line 182:
```suggestion
      return chapter?.text ?? '';
```

src/plugins/chapters/index.js
```diff
@@ -0,0 +1,195 @@
...
+      const originalUpdateFn = Object.getPrototypeOf(this).update;
+      originalUpdateFn.call(this, seekBarRect, seekBarPoint, content);
+      chapterEl.innerHTML = getChapterFromPoint(seekBarPoint);
+    };
+  };
+
+  return ChaptersPlugin;
+})();
+
+export default chapters;
```
ehab-cl, line 195: I understood that the plugin code is probably a template that gets filled in, it is worth checking if we could have a more modern template (Using class instead of prototype and using arrow functions instead of regular functions) .

src/plugins/vtt-thumbnails/index.js
```diff
@@ -89,179 +85,65 @@ const VttThumbnailsPlugin = (function () {
...
+    const halfThumbnailWidth = parseInt(this.vttData[0].css.width, 10) / 2;
+
+    const updateThumbnailStyle = this.updateThumbnailStyle.bind(this);
+
+    this.player.getDescendant([
+      'controlBar',
+      'progressControl',
+      'seekBar',
+      'mouseTimeDisplay'
+    ]).update = function (seekBarRect, seekBarPoint) {
```
ehab-cl, line 130:
```suggestion
     const mouseTimeDisplay = this.player.getDescendant([
      'controlBar',
      'progressControl',
      'seekBar',
      'mouseTimeDisplay'
    ]);
    mouseTimeDisplay.update = function (seekBarRect, seekBarPoint) {
```

src/plugins/vtt-thumbnails/index.js
```diff
@@ -89,179 +85,65 @@ const VttThumbnailsPlugin = (function () {
...
+      const width = seekBarRect.width;
+
+      updateThumbnailStyle(time);
+
+      this.getChild('timeTooltip').updateTime(seekBarRect, seekBarPoint, time, () => {
+        if ((seekBarRect.width * seekBarPoint) < halfThumbnailWidth) {
+          this.el_.style.left = halfThumbnailWidth;
+        } else if ((seekBarRect.width * seekBarPoint) + halfThumbnailWidth > width) {
+          this.el_.style.left = (seekBarRect.width * seekBarPoint) - halfThumbnailWidth;
+        } else {
```
ehab-cl, line 141: it could be because i am not familiar with the codebase, but i am having a hard time understand why this specific positioning, for that reason i would add comments about why this is needed

src/video-player.utils.js
```diff
@@ -127,4 +126,10 @@ export const overrideDefaultVideojsComponents = () => {
     // Position the 'logo-button' button last
     children.push('logoButton');
   }
+
+  const SeekBar = videojs.getComponent('SeekBar');
+  if (SeekBar && !SeekBar.prototype.options_.children.includes('mouseTimeDisplay')) {
+    SeekBar.prototype.options_.children.splice(1, 0, 'mouseTimeDisplay');
+  }
```
ehab-cl, line 133: what is this doing? a comment would really help

ehab-cl (approved):
LGTM 💪

src/video-player.utils.js
tsi: videojs team made a decision to remove this component on mobile devices. we want it there since we're attaching the chapter display to it. I'll add a comment.

src/plugins/chapters/index.js
tsi: Yeah, not really a template, it all started with the seek-thumbnails plugin that was copied as-is from somewhere.
I realize it's not modern, but it's actually pretty clear, readable and well structured which is what I find most important.
Switching to a Class would be easy, but if I would re-write these I'd go the functional route, getting rid of all these `this` :)

src/plugins/vtt-thumbnails/index.js
tsi: we want it to stop "moving" before the thumbnails exits the player which is at `halfThumbnailWidth` distance from the edge.

tsi: Added a comment

tsi:
@ehab-cl re:

> Chapters can be configured in two ways:

I think it's really nice-to-have for our users to be able to programmatically setup chapters without having to store a VTT file somewhere.
Regarding non-standard - under the hood it is programmatically doing the exact same thing - i.e. calling `addTextTrack` instead of `addRemoteTextTrack`

src/plugins/chapters/index.js
tsi: eslint didn't like that

ehab-cl (approved):
💪

PR #443, opened 2023-10-10
fix: re-run analytics also on cld source change

jakub-roch:
Run analytics also on cld analytics change - not only basic source change.

### DIFF
--- a/src/plugins/cloudinary-analytics/index.js
+++ b/src/plugins/cloudinary-analytics/index.js
@@ -16,19 +16,22 @@ class CloudinaryAnalytics {
     publicId: this.player.cloudinary.currentPublicId()
   })
 
+  sourceChanged = () => {
+    const metadata = this.getMetadata();
+    if (metadata.cloudName && metadata.publicId) {
+      this.currentVideMetadata = metadata;
+      this.cloudinaryAnalytics.startManualTracking(metadata, {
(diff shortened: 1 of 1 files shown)


src/plugins/cloudinary-analytics/index.js
```diff
@@ -16,19 +16,22 @@ class CloudinaryAnalytics {
...
-        this.cloudinaryAnalytics.startManualTracking(metadata, {
-          videoPlayerType: 'cloudinary video player',
-          videoPlayerVersion: VERSION
-        });
-      } else if (this.currentVideMetadata.cloudName !== metadata.cloudName || this.currentVideMetadata.publicId !== metadata.publicId) {
-        this.cloudinaryAnalytics.stopManualTracking();
-      }
-    });
+    this.player.on(PLAYER_EVENT.CLD_SOURCE_CHANGED, this.sourceChanged);
+    this.player.on(PLAYER_EVENT.SOURCE_CHANGED, this.sourceChanged);
```
ehab-cl, line 34: Are we sure that these 2 events are not overlapping? that is to say that it is always one of them and never both?

ehab-cl (approved):
💪

src/plugins/cloudinary-analytics/index.js
jakub-roch: I checked that only one of them is fired

PR #444, opened 2023-10-11
ME-13953-ads-fix

tsi:


docs/vast-vpaid.html
```diff
@@ -37,20 +34,13 @@
         cloud_name: 'demo',
         ads: {
           adTagUrl: adTagUrl,
+          debug: true
```
ehab-cl, line 37: should not we remove this now?

ehab-cl (approved):
💪

docs/vast-vpaid.html
tsi: I'm now making this configurable, instead of always `true`.
I don't mind it will be `true` in the example page

PR #447, opened 2023-10-12
ME-13952-chapters-button

tsi:


docs/chapters.html
```diff
@@ -30,7 +30,6 @@
       const playerVTT = cloudinary.videoPlayer('player-vtt', {
         cloud_name: 'demo',
         publicId: 'snow_horses',
-        info: { title: 'Snow Horses', subtitle: 'A movie about horses' },
```
ehab-cl, line 33: I think its a good idea to keep one of them, to see how chapters would work with title and sub title.

src/plugins/chapters/index.js
```diff
@@ -112,7 +112,9 @@ const ChaptersPlugin = (function () {
       });
       this.chaptersTrack = textTrack.track;
       this.setupChaptersDisplays();
-      this.player.controlBar.chaptersButton.update();
+      if (this.player.controlBar.chaptersButton) {
+        this.player.controlBar.chaptersButton.update();
+      }
```
ehab-cl, line 117: you can just do 

```
this.player.controlBar?.chaptersButton?.update();
```

PR #448, opened 2023-10-17
ME-14018-playlist-chapters

tsi:
Fix chapters interface when switching video sources

### DIFF
--- a/src/video-player.js
+++ b/src/video-player.js
@@ -322,8 +322,10 @@ class VideoPlayer extends Utils.mixin(Eventable) {
 
   _initChapters() {
     this.videojs.on(PLAYER_EVENT.CLD_SOURCE_CHANGED, (e, { source }) => {
-      if (!isEmpty(source._chapters)) {
-        this.videojs.chapters(source._chapters);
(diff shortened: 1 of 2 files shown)


src/video-player.js
```diff
@@ -322,8 +322,10 @@ class VideoPlayer extends Utils.mixin(Eventable) {
 
   _initChapters() {
     this.videojs.on(PLAYER_EVENT.CLD_SOURCE_CHANGED, (e, { source }) => {
-      if (!isEmpty(source._chapters)) {
-        this.videojs.chapters(source._chapters);
+      if (!isEmpty(source._chapters) && this.videojs.chapters) {
+        isFunction(this.videojs.chapters)
+          ? this.videojs.chapters(source._chapters)
+          : this.videojs.chapters.src(source._chapters);
```
ehab-cl, line 328: It looks a bit complex to me, is this changing behaviour from videojs or from our code?

ehab-cl (approved):
💪

src/video-player.js
tsi: It's just that the plugin needs to be called differently on init and on source update. Same for our other custom plugins. We can probably change that.

PR #453, opened 2023-10-25
ME-13973-default-f_auto

tsi:


src/plugins/cloudinary/common.js
```diff
@@ -204,6 +204,21 @@ const codecToSrcTransformation = codec => {
   }
 };
 
+const setupCloudinaryMiddleware = () => {
+  // Allow 'auto' as a source type
+  videojs.use('video/auto', () => {
+    return {
+      async setSource(srcObj, next) {
+        const headers = await fetch(srcObj.src, { method: 'HEAD' }).then(r => r.headers);
```
ehab-cl, line 212: I think the following is easier to read 

```
const { headers } = await fetch(srcObj.src, { method: 'HEAD' })
```

src/plugins/cloudinary/common.js
```diff
@@ -204,6 +204,21 @@ const codecToSrcTransformation = codec => {
...
+
+        return next(null, {
```
ehab-cl, line 214: I am not familiar at all with how videjs middleware works, but i assume it is something similar to express js, in this case it is worth checking how this code behaves in the following cases - in case you already did not:

- Updating source 
- Playlist
- When for whatever reason the promise rejects for now you are not calling next at all when the promise rejects, so what does this mean if this happened for example in a playlist

ehab-cl (approved):
💪

src/plugins/cloudinary/common.js
tsi: Sure, I tested playlists and those work fine.
I'll try to look into what can be done with failures, I guess you'll just get some media-not-found error.

tsi: Yeah, so if for whatever reason the `fetch` fails, you'll get this

image

PR #458, opened 2023-11-28
ME-14240-sp_auto-default

tsi:
This PR changes the default `streaming_profile` for ABR streams from `none` to `auto`

### DIFF
--- a/src/plugins/cloudinary/models/video-source/video-source.js
+++ b/src/plugins/cloudinary/models/video-source/video-source.js
@@ -4,6 +4,7 @@ import { assign } from 'utils/assign';
 import { objectToQuerystring } from 'utils/querystring';
 import { castArray } from 'utils/array';
 import { SOURCE_TYPE } from 'utils/consts';
+import { isKeyPresent } from 'utils/object';
 import {
   CONTAINER_MIME_TYPES,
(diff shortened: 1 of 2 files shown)


src/plugins/cloudinary/models/video-source/video-source.js
```diff
@@ -217,6 +218,14 @@ class VideoSource extends BaseSource {
         });
       }
 
+      if (isAdaptive) {
+        if (!isKeyPresent(opts.transformation, 'streaming_profile')) {
```
ehab-cl, line 222: I think the custom implementation of isKeyPresent is an overkill, you could if you want use `has` from lodash or simply do - unless i am missing something here 
```
if (opts?.transformation?.streaming_profile !== undefined) {
}
```

tsi: Yeah, I wasn't happily going that route.
The thing is that transformations can come in many shapes - they can be a flat object, a nested object or an array of objects and so on.
We don't know where and how `streaming_profile` will be defined, but we need to find if it is.
The only other way I could think of is stringifying the whole thing and search for the `"streaming_profile":` string.

ehab-cl (approved):
💪

PR #460, opened 2023-12-04
ME-14242-internal-analytics-fix

tsi:
cast `analytics` to a boolean


ehab-cl:
What about not removing defaults?

PR #464, opened 2023-12-11
ME-14565-subtitles-styles

tsi:


src/plugins/styled-text-tracks/index.js
```diff
@@ -0,0 +1,9 @@
+export default async function lazyPlugin(config) {
+  const player = this;
+  try {
+    const { default: initPlugin } = await import('./styled-text-tracks');
```
ehab-cl, line 4: 💪

src/plugins/styled-text-tracks/styled-text-tracks.scss
```diff
@@ -0,0 +1,71 @@
+.cld-video-player {
+  // Positions
+  // Center
```
ehab-cl, line 3: I would say the positions are pretty clear from the class names (if you agree with me applies to the following classes as well, including themes)
```suggestion
```

ehab-cl (approved):
LGTM 💪

src/plugins/styled-text-tracks/styled-text-tracks.scss
tsi: Yeah, I was using them as more of separators rather than to clarify meaning. Removed.

ehab-cl (approved):

PR #465, opened 2023-12-11
ME-14358-paced-transcript-event

tsi:
This change fixes an issue that came up in QA where the textTrack was not always being added because of a race-condition with the video source


ehab-cl (approved):
LGTM 💪

PR #467, opened 2023-12-14
ME-14358-paced-transcript-safari

tsi:
Really wanted to find a better solution to this

### DIFF
--- a/src/plugins/paced-transcript/index.js
+++ b/src/plugins/paced-transcript/index.js
@@ -31,11 +31,13 @@ function pacedTranscript(config) {
         mode: options.default ? 'showing' : 'disabled'
       });
 
-      setTimeout(() => { // required for Safari to display the captions
-        captions.forEach(caption => {
-          captionsTrack.track.addCue(new VTTCue(caption.startTime, caption.endTime, caption.text));
-        });
-      }, 100);
+      // required for Safari to display the captions
+      // https://github.com/videojs/video.js/issues/8519
+      await new Promise(resolve => setTimeout(resolve, 100));
+
+      captions.forEach(caption => {
+        captionsTrack.track.addCue(new VTTCue(caption.startTime, caption.endTime, caption.text));
+      });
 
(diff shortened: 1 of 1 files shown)


src/plugins/paced-transcript/index.js
```diff
@@ -31,9 +31,12 @@ function pacedTranscript(config) {
         mode: options.default ? 'showing' : 'disabled'
       });
 
-      captions.forEach(caption => {
-        captionsTrack.track.addCue(new VTTCue(caption.startTime, caption.endTime, caption.text));
-      });
+      setTimeout(() => { // required for Safari to display the captions
+        captions.forEach(caption => {
+          captionsTrack.track.addCue(new VTTCue(caption.startTime, caption.endTime, caption.text));
```
ehab-cl, line 36: I have two suggestions here, but honestly no idea if they will actually work, but  if they work i think they will be easier to read:

- It seems [here](https://videojs.com/guides/text-tracks/#:~:text=Note%3A%20If%20you%20need%20a%20callback%2C%20instead%20of%20a%20callback%20you%20could%20use%20the%20technique%20below%3A) that addRemoteTextTrack has a load event, the example here is about loading a src file, in your case you are adding a text track directly, but still that might work. 

- instead of setTimeout maybe you can try waiting for a dummy promise just like 

```
const captionsTrack = await player.addRemoteTextTrack({
```

ehab-cl (approved):
💪

src/plugins/paced-transcript/index.js
tsi: So this specifically [won't work](https://codesandbox.io/s/minimal-videojs-texttracks-await-d69fcw?file=/index.html) but it's a good direction as adding a "sleep" Promise [does work](https://codesandbox.io/s/minimal-videojs-texttracks-sleep-promise-s9h6rp?file=/index.html:1211-1266) and looks much better then the `setTimeout` wrapper

PR #468, opened 2023-12-14
ME-6001-vp-wave-a11y

tsi:
image

### DIFF
--- a/src/video-player.utils.js
+++ b/src/video-player.utils.js
@@ -125,6 +125,10 @@ export const overrideDefaultVideojsComponents = () => {
 
     // Position the 'logo-button' button last
     children.push('logoButton');
+
+    // Remove these button
+    children.splice(children.indexOf('skipForward'), 1);
+    children.splice(children.indexOf('skipBackward'), 1);
   }
 
   const SeekBar = videojs.getComponent('SeekBar');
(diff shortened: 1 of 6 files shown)


src/video-player.utils.js
```diff
@@ -125,6 +125,10 @@ export const overrideDefaultVideojsComponents = () => {
 
     // Position the 'logo-button' button last
     children.push('logoButton');
+
+    // Remove these button
+    children.splice(children.indexOf('skipForward'), 1);
+    children.splice(children.indexOf('skipBackward'), 1);
```
ehab-cl, line 131: were these basically not visible but still part of the DOM? i am trying to understand why is being done here exactly?

ehab-cl (approved):
LGTM 💪

src/video-player.utils.js
tsi: These are [videojs components](https://docs.videojs.com/control-bar_skip-buttons_skip-forward.js.html) that if not configured, still render empty, we don't need them currently because we have our own implementation of skip buttons.
Two things I noted to myself 😄 :
1. Fix the videojs accessibility issue, I'll create a PR when I get to it
2. Our implementation is redundant, we can think of removing it

PR #469, opened 2023-12-18
ME-14565-subtitles-styles-gravity

tsi:
This PR adjust the styled-subtitles positioning method to be:
```
        options: {
          theme: "", // one of 'default', 'videojs-default', 'yellow-outlined', 'player-colors' & '3d'
          fontFace: "", // any Google font
          fontSize: "", // any CSS value
          gravity: "", // i.e. 'top-left', 'center' etc
          box: { // Object of x/y/width/height
            x: "0%",
            y: "0%",
            width: "100%",
            height: "100%"
          },
          style: { // Styles key-value object
            color: "hotpink"
          }
        },
```

### DIFF
--- a/docs/subtitles-and-captions.html
+++ b/docs/subtitles-and-captions.html
@@ -127,13 +127,15 @@
       const textTracks = {
         options: {
           // theme: "", // one of 'default', 'videojs-default', 'yellow-outlined', 'player-colors' & '3d'
-          // position: "", // i.e. 'top-left', 'center' etc
-          // offset: { // Object of top/right/bottom/left
-          //   top: "25%",
-          //   left: "12%"
-          // },
           // fontFace: "", // any Google font
           // fontSize: "", // any CSS value
+          // gravity: "", // i.e. 'top-left', 'center' etc
+          // box: { // Object of x/y/width/height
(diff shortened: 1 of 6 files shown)


docs/subtitles-and-captions.html
```diff
@@ -388,13 +396,15 @@ <h3 class="mt-4">Example Code:</h3>
...
-            //   left: "12%"
-            // },
             // fontFace: "", // any Google font
             // fontSize: "", // any CSS value
+            // gravity: "", // i.e. 'top-left', 'center' etc
+            // box: { // Object of x/y/width/height
+            //   x: "0%",
+            //   y: "0%",
+            //   width: "100%",
+            //   height: "100%"
```
ehab-cl, line 406: what is the usage of declaring width, height? i mean what use case it solves for you, does it provide an over-flowing responsive container?

ehab-cl (approved):
💪

docs/subtitles-and-captions.html
tsi: It allows you to shift the box using x/y while making it smaller to make sure it doesn't overflow

PR #470, opened 2023-12-19
fix: remove cache VTT chapters

jakub-roch:


ehab-cl (approved):
Approving this a temporary solution so that we release it before the freeze and after it we should think of a more scalable approach.

PR #502, opened 2024-01-01
fix: deploy workflows (ME-4402)

tsi:


.github/workflows/unpublish.yml
```diff
@@ -0,0 +1,39 @@
+name: ⛔ Deprecate
```
ehab-cl, line 1: File name does not match workflow name, it is better to align them, in this case i believe it is better to change workflow file to be deprecate.yml since indeed this is about deprecating and not "un-publish"

.github/workflows/unpublish.yml
```diff
@@ -0,0 +1,39 @@
...
+
+    steps:
+    - name: Setup Node.js
+      uses: actions/setup-node@v4
+      with:
+        node-version: '20.x'
+        registry-url: 'https://registry.npmjs.org'
+
+    - name: Deprecate version
+      run: npm deprecate cloudinary-video-player@${{ github.event.inputs.version }} "This version of cloudinary-video-player has been deprecated."
```
ehab-cl, line 32: I really think we rarely if ever want to deprecate a version. I am not sure if this is also how you think of it as well, but regarding this workflow we need to make sure that it is clear this is not about reverting or rolling back a version this is about deprecating a version. 

That being said if you want now to just make latest point a to a previous version (rollback), how do you do that?

package.json
```diff
@@ -59,7 +56,7 @@
       },
       {
         "path": "./dist/cld-video-player.light.min.js",
-        "maxSize": "145kb"
+        "maxSize": "150kb"
```
ehab-cl, line 59: i am just curious why we had to change size here? i mean all your changes seems dev changes?

src/plugins/index.js
```diff
@@ -7,38 +7,39 @@ import interactive from './interactive-plugin';
...
+import contextMenu from './context-menu';
+import floatingPlayer from './floating-player';
 import pacedTranscript from './paced-transcript';
+import vttThumbnails from './vtt-thumbnails';
 
+// Lazy loaded plugins
 import styledTextTracks from './styled-text-tracks';
 
 const plugins = {
+  // #if (!process.env.WEBPACK_BUILD_LIGHT)
```
ehab-cl, line 26: not sure i understand why do we have these changes?

ehab-cl (approved):
LGTM, i have left few comments and suggestions

.github/workflows/release-edge.yml
```diff
@@ -0,0 +1,70 @@
...
+          registry-url: 'https://registry.npmjs.org'
+        if: ${{ steps.release.outputs.release_created }}
+        # these if statements ensure that a publication only occurs when
+        # a new release is created:
+      - run: npm run build-all
+        if: ${{ steps.release.outputs.release_created }}
+      - run: npm publish --tag edge
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+        if: ${{ steps.release.outputs.release_created }}
```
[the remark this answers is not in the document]
tsi: I need the condition on every step I want to run conditionally.
I did try to extract the whole `publish` block into a composite action but it created a lot of added complexity

.github/workflows/release-edge.yml
```diff
@@ -0,0 +1,70 @@
...
+
+      - name: Slack Notification - Deployment Success
+        if: steps.release.outputs.release_created && success()
+        uses: rtCamp/action-slack-notify@v2.2.0
+        env:
+          SLACK_WEBHOOK: ${{ vars.FE_DEPLOYMENTS_SLACK_WEBHOOK }}
+          SLACK_CHANNEL: 'rnd-fe-releases'
+          SLACK_COLOR: ${{ job.status }}
+          SLACK_TITLE: 'Video Player ${{ steps.release.outputs.tag_name }} Deployed'
+          SLACK_MESSAGE: 'Success :rocket: cloudinary-video-player version ${{ steps.release.outputs.tag_name }} deployed successfully'
```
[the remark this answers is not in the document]
tsi: yes, but the `tag_name` would show if this is `edge` or not, also, I tend to forward these to the VP slack channel only for stable releases

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20.x'
+          registry-url: 'https://registry.npmjs.org'
+        if: ${{ steps.release.outputs.release_created }}
+      - run: npm ci
+        if: ${{ steps.release.outputs.release_created }}
+      - run: npm run build-all
+        if: ${{ steps.release.outputs.release_created }}
+      - run: npm publish
```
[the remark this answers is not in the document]
tsi: That's correct, the `if` distinguishes the change PR and the release PR.
You merge a change PR, this trigger release-please which creates a release PR, only when this PR is merged this `if` becomes truthy and a release gets created

.github/workflows/unpublish.yml
tsi: This workflow _is_ about rolling-back/unpublishing.
Unfortunately since the left-pad incident you can not unpublish a package if you have dependant packages.
Our best option, if we have a faulty version is to deprecate it and set the latest/edge tag to the previous valid version, and that's what this workflow is doing

package.json
tsi: I was surprised as well, I guess some infra/backend caused the build to slightly increase in size (it was a very small increase)

src/plugins/index.js
tsi: You're right, apologies :)
Some commits here were made just to trigger a build, I could have cleaned it up but they were nice cleanups, here for example I sorted the plugins alphabetically

.github/actions/calculate-edge-version/index.js
```diff
@@ -0,0 +1,15 @@
+const core = require('@actions/core');
+const semver = require('semver');
+
+try {
+  let packageVersion = core.getInput('version', { require: true });
+
+  // Remove initial v if present
+  if (packageVersion.charAt(0) === 'v') {
+    packageVersion = packageVersion.substring(1);
+  }
```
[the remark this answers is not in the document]
tsi: I can go with this one:
```suggestion
  let packageVersion = core.getInput('version', { require: true }).replace(/^v/, ''); // Remove initial v 
```
IMHO the extra variable + the ternary operation are not easier to read

.github/actions/fetch-latest-release/index.js
```diff
@@ -0,0 +1,65 @@
...
+
+  core.info(`Fetching the latest release for \`${owner}/${repo}\``)
+
+  try {
+    latestRelease = await octokit.rest.repos.getLatestRelease({
+      owner,
+      repo,
+    })
+  } catch (error) {
+    core.info('Could not fetch the latest release. Have you made one yet?')
```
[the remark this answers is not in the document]
tsi: In the output log

.github/actions/calculate-edge-version/index.js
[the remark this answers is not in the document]
ehab-cl: @tsi your call, if i have to choose between what you haev now and the regex i will chose what you have now 😄 
I hate regex

.github/workflows/release-edge.yml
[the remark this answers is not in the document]
ehab-cl: yea i missed the run command that was before it 😄

.github/workflows/unpublish.yml
ehab-cl: @tsi what is the left pad incident? 
Anyway i am bit afraid of this workflow, why can not we have something that changes where the latest tag points to? also deprecating usually happens in favor of a new version but here we are saying it happens in favor of an old version, not sure how this eventually will play out?

tsi: https://qz.com/646467/how-one-programmer-broke-the-internet-by-deleting-a-tiny-piece-of-code 😄 

Note that this workflow does 2 things:
1. `npm deprecate`
2. `npm dist-tag` which is doing what you said - moving where the latest/edge tag points to
the `deprecate` part is just a "bonus" to let future users know that this version should not be used.

tsi:
Moved same changes to PR 507

PR #517, opened 2024-01-15
feat: profiles with default profiles

jakub-roch:
Add profiles first version for testing purpose.
Whole feature is available via property `__profileName` which is not production-ready yet

ONLY FOR TESTING PURPOSE - IF PROFILE IS NOT PROVIDED EVERYTHING ELSE SHOULD WORK AS PREVIOUSLY

### DIFF
--- a/docs/profiles.html
+++ b/docs/profiles.html
@@ -1,0 +1,129 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <title>Cloudinary Video Player</title>
+  <link href="https://res.cloudinary.com/cloudinary-marketing/image/upload/f_auto,q_auto/c_scale,w_32/v1597183771/creative_staging/cloudinary_internal/Website/Brand%20Updates/Favicon/cloudinary_web_favicon_192x192.png" rel="icon" type="image/png">
+
+  <!-- Bootstrap -->
+  <link href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
+
+  <!-- highlight.js -->
+  <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/styles/solarized-light.min.css">
+  <script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/highlight.min.js"></script>
+  <script>hljs.initHighlightingOnLoad();</script>
+
+  <!--
+    We're loading scripts & style dynamically for development/testing.
+    Real-world usage would look like this:
+
+    <link rel="stylesheet" href="https://unpkg.com/cloudinary-video-player/dist/cld-video-player.min.css">
+    <script src="https://unpkg.com/cloudinary-video-player/dist/cld-video-player.min.js"></script>
+  -->
+
+  <script type="text/javascript" src="./scripts.js"></script>
+
+  <script type="text/javascript">
+    window.addEventListener('load', function() {
+      const playerWithDefaultProfile = cloudinary.videoPlayer('player-default-profile', {
+        cloud_name: 'demo',
+        __profileName: 'defaultProfile1',
+      }, function () {
+        playerWithDefaultProfile.source('sea_turtle');
+      });
+
+      const playerWithCustomProfile = cloudinary.videoPlayer('player-custom-profile', {
+        cloud_name: 'dwaq5xqm4',
+        __profileName: 'default_profile.json',
+      }, function () {
+        playerWithCustomProfile.source('samples/cld-sample-video');
+      });
+    }, false);
+  </script>
+</head>
+<body>
+<div class="container p-4 col-12 col-md-9 col-xl-6">
+  <nav class="nav mb-2">
+    <a href="./index.html">&#60;&#60; Back to examples index</a>
+  </nav>
+  <h1>Cloudinary Video Player</h1>
+  <h3 class="mb-4">Profiles</h3>
+
+  <p>
+    Profiles are still under development and it's not safe for use on production.<br/>
+    For testing purpose you can use Profiles Feature according to the instructions below:
+    <br/><br/>
+    1. Provide `__profileName` config property using one of default profiles (`defaultProfile1`, `defaultProfile2`)<br/>
+    2. You can use your own profile JSON, create folder on your cloud named `profiles` and upload json file
+    - once you have it just provide publicId with extension of your uploaded file as value of `__profileName`
+  </p>
+
+
+  <h5>Player with default profile</h5>
+  <video
+    id="player-default-profile"
+    playsinline
+    controls
+    autoplay
+    muted
+    class="cld-video-player"
+    width="500">
+  </video>
+
+  <br/><br/>
+  <h5>Player with custom profile</h5>
+  <video
+    id="player-custom-profile"
+    playsinline
+    controls
+    autoplay
+    muted
+    class="cld-video-player"
+    width="500">
+  </video>
+
+  <h3 class="mt-4">Example Code:</h3>
+
+  <pre><code class="language-html">
+
+      &lt;video
+        id="player-default-profile"
+        controls
+        autoplay
+        muted
+        class="cld-video-player"
+        width="500"&gt;
+      &lt;/video&gt;
+
+    &lt;video
+        id="player-custom-profile"
+        controls
+        autoplay
+        muted
+        class="cld-video-player"
+        width="500"&gt;
+      &lt;/video&gt;
+
+      </code>
+      <code class="language-javascript">
+        // default profile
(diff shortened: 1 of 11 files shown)


src/components/profiles/profiles.js
```diff
@@ -0,0 +1,44 @@
...
+import { fetchWithTimeout } from '../../utils/fetch-with-timeout';
+
+const defaultProfilesNames = ['loopingVideo', 'videoStream', 'stylish'];
+
+export const getDefaultProfile = async (profileName) => {
+  const defaultProfiles = {
+    loopingVideo: await import(/* webpackChunkName: "profiles/loopingVideo" */ '../../profiles/loopingVideo.json'),
+    videoStream: await import(/* webpackChunkName: "profiles/videoStream" */ '../../profiles/videoStream.json'),
+    stylish: await import(/* webpackChunkName: "profiles/stylish" */ '../../profiles/stylish.json')
+  };
```
tsi, line 11: If you load this whole module lazily you wouldn't need to do that

webpack/build-utils.js
```diff
@@ -6,4 +8,9 @@ const minFilenamePart = isMin ? '.min' : '';
 
 const lightFilenamePart = isLight ? '.light' : '';
 
-module.exports = { isMin, isLight, minFilenamePart, lightFilenamePart };
+const getProfilesPathPattern = (outputPath) => ({
+  from: path.resolve(outputPath, '../src/profiles'),
+  to: path.resolve(outputPath, 'profiles')
+});
```
tsi, line 14: why did you need this webpack thing? can’t package.json exports handle that?

src/profiles/_profileSupportedOptions.json
```diff
@@ -0,0 +1,72 @@
+{
```
tsi, line 1: what do you plan to do with that list? it's not being imported anywhere

docs/profiles.html
```diff
@@ -0,0 +1,129 @@
...
+
+  <script type="text/javascript" src="./scripts.js"></script>
+
+  <script type="text/javascript">
+    window.addEventListener('load', function() {
+      const playerWithDefaultProfile = cloudinary.videoPlayer('player-default-profile', {
+        cloud_name: 'demo',
+        __profileName: 'loopingVideo',
+      }, function () {
+        playerWithDefaultProfile.source('sea_turtle');
```
tsi, line 32: What is this 3rd argument of `videoPlayer()`? does that mean we will not be able to call the `source()` method normally?

src/video-player.js
```diff
@@ -514,7 +556,8 @@ class VideoPlayer extends Utils.mixin(Eventable) {
   }
 
   source(publicId, options = {}) {
-    ({ publicId, options } = normalizeOptions(publicId, options));
+    const extendedOptions = Utils.assign({}, this.profileOptions.sourceOptions, options);
```
tsi, line 559: Should not be done here, not all players go through here.
If you do this in the `cloudinary` plugin `source` method you can be sure it will be called on every souce-change

src/video-player.js
```diff
@@ -24,6 +24,7 @@ import { extendCloudinaryConfig, normalizeOptions, isRawUrl } from './plugins/cl
 // #if (!process.env.WEBPACK_BUILD_LIGHT)
 import qualitySelector from './components/qualitySelector/qualitySelector.js';
 import { interactionAreaService } from './components/interaction-area/interaction-area.service';
+import { getProfile } from './components/profiles/profiles';
```
tsi, line 27: this means it will not be available in the light build, was that the intention?

src/video-player.js
```diff
@@ -564,7 +607,7 @@ class VideoPlayer extends Utils.mixin(Eventable) {
   }
 
   playlist(sources, options = {}) {
-    options = Utils.assign({}, options, { playlistWidget: this.playerOptions.playlistWidget });
+    options = Utils.assign({}, this.profileOptions.playlistOptions, options, { playlistWidget: this.playerOptions.playlistWidget });
```
tsi, line 610: see previous comment

src/video-player.js
```diff
@@ -575,7 +618,7 @@ class VideoPlayer extends Utils.mixin(Eventable) {
   }
 
   playlistByTag(tag, options = {}) {
-    options = Utils.assign({}, options, { playlistWidget: this.playerOptions.playlistWidget });
+    options = Utils.assign({}, this.profileOptions.playlistByTagOptions, options, { playlistWidget: this.playerOptions.playlistWidget });
```
tsi, line 621: see previous comment

src/video-player.js
```diff
@@ -587,6 +630,7 @@ class VideoPlayer extends Utils.mixin(Eventable) {
   }
 
   sourcesByTag(tag, options = {}) {
+    options = Utils.assign({}, this.profileOptions.sourcesByTagOptions, options);
```
tsi, line 633: see previous comment

src/video-player.js
```diff
@@ -107,6 +130,25 @@ class VideoPlayer extends Utils.mixin(Eventable) {
     this._setVideoJsListeners(ready);
   }
 
+  initProfile = async function (profileName, initOptions) {
+    return getProfile(initOptions.cloud_name, profileName, 10000)
```
tsi, line 134: Could be nicer to use `await` here

src/video-player.js
```diff
@@ -61,16 +62,38 @@ class VideoPlayer extends Utils.mixin(Eventable) {
...
+  initPlayer(initOptions, ready) {
+    // create empty profile options for simple extends
+    if (!this.profileOptions) {
+      this.profileOptions = {
+        playerOptions: {},
+        sourceOptions: {},
+        playlistOptions: {},
+        playlistByTagOptions: {},
+        sourcesByTagOptions: {}
+      };
```
tsi, line 91: I'd prefer if we wouldn't need to do that. preparing empty placeholders for everything even if most won't need it is just adding noise.

tsi (commented):
That's awesome, please see my comments

webpack/build-utils.js
jakub-roch: I mean by that we can have raw json files instead of js files that exports json objects

src/components/profiles/profiles.js
jakub-roch: good suggestion, changed it 👍

src/profiles/_profileSupportedOptions.json
jakub-roch: its being used but wanted to keep it as some official list that can be used for other layers like backend

src/video-player.js
jakub-roch: changed to be lazy loaded so not relevant

src/video-player.js
jakub-roch: 👍

src/components/profiles/profiles.js
```diff
@@ -0,0 +1,48 @@
...
+  return await response.json();
+};
+
+export const getProfile = async (cloudName, profileName, timeout = 10000) => {
+  if (defaultProfilesNames.includes(profileName)) {
+    return await getDefaultProfile(profileName);
+  }
+
+  return await getCustomProfile(cloudName, profileName, timeout);
+};
```
tsi, line 48: @jakub-roch a component should usually return some UI
I'd say it's a plugin, or a utility, but it shouldn't be a component

src/profiles/_profileSupportedOptions.json
tsi: lists gets outdated and need maintenance, I wouldn't do that unless it's strictly required

src/profiles/loopingVideo.json
```diff
@@ -0,0 +1,15 @@
+{
```
tsi, line 1: Can we put these profiles next to [the default](https://github.com/cloudinary/cloudinary-video-player/blob/edge/src/config/defaults.js) one?

src/video-player.js
```diff
@@ -61,16 +61,38 @@ class VideoPlayer extends Utils.mixin(Eventable) {
...
-    this._videojsOptions = this.options.videojsOptions;
-
     // Make sure to add 'video-js' class before creating videojs instance
     this.videoElement.classList.add('video-js');
 
+    // first phase - shouldn't be exposed to people until we have BE ready
+    const profileName = initOptions.__profileName;
+
+    if (profileName) {
+      this.initProfile(profileName, initOptions)
```
tsi, line 73: Why passing `profileName, initOptions` when profileName is a member of initOptions?

Also you should probably pass this.options, and not just initOptions since initOptions doesn't go through extractOptions

src/components/profiles/profiles.js
tsi: Besides that, this can be much simplified, will something like this work?

```
import { getCloudinaryUrl } from "plugins/cloudinary/common";

import * as profiles from "config";

const getCustomProfile = async (cloudName, profileName) => {
  const profileUrl = getCloudinaryUrl(`profiles/${profileName}`, {
    cloud_name: cloudName,
    resource_type: "raw",
    secure: true,
  });

  return fetch(profileUrl, {
    method: "GET"
  });
};

export const getProfile = async (cloudName, profileName) => {
  if (profiles[profileName]) {
    return profiles[profileName];
  }

  return await getCustomProfile(cloudName, profileName);
};
```

PR #518, opened 2024-01-21
ME-14384-lazy-recommendations

tsi:
This PR wraps the Recommendations component with a lazy loading wrapper, to be fetched only when required

### DIFF
--- a/src/components/recommendations-overlay/index.js
+++ b/src/components/recommendations-overlay/index.js
@@ -1,3 +1,11 @@
-import RecommendationsOverlay from './recommendations-overlay';
-
-export default RecommendationsOverlay;
+export default async function lazyComponent(player) {
+  try {
+    if (!player.getChild('recommendationsOverlay')) {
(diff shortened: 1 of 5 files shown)


src/components/recommendations-overlay/index.js
```diff
@@ -1,3 +1,11 @@
-import RecommendationsOverlay from './recommendations-overlay';
-
-export default RecommendationsOverlay;
+export default async function lazyComponent(player) {
```
ehab-cl, line 1: This will make things easier for global search and for readability 
```suggestion
export default async function lazyRecommendationsOverlay(player) {
```

ehab-cl (dismissed):
💪

src/components/recommendations-overlay/index.js
tsi: 👍 
only that it's a component, not a plugin :)

jakub-roch (approved):
Looks good 👍

PR #519, opened 2024-01-22
ME-14382-lazy-shoppable-plugin

tsi:
This PR wraps the shoppable plugin with a lazy loading wrapper


src/plugins/shoppable-plugin/index.js
```diff
@@ -0,0 +1,6 @@
+export default async function lazyPlugin(player, options) {
```
ehab-cl, line 1:
```suggestion
export default async function lazyShoppablePlugin(player, options) {
```

src/video-player.js
```diff
@@ -533,9 +533,8 @@ class VideoPlayer extends Utils.mixin(Eventable) {
       return this.videojs.cloudinary.source(publicId, options);
     }
 
-    // Interactive plugin - available in full (not light) build only
-    if (this.videojs.interactive) {
-      this.videojs.interactive(this.videojs, options);
```
ehab-cl, line 538: why are we removing these 2 lines?

ehab-cl (dismissed):
💪

src/video-player.js
tsi: The "interactive" plugin was supposed to wrap the shoppable plugin as well as other relevant features, this never happened and it only contains the shoppable plugin.
Meanwhile, the interaction-area plugin was introduced, so this just became even more confusing.
Bottom line - the "interactive" plugin became "shoppable"

jakub-roch (approved):

PR #522, opened 2024-01-24
ME-14916-picture-in-picture

tsi:


src/config/defaults.js
```diff
@@ -9,10 +9,8 @@ export default {
   playsinline: videojs.browser.IS_IOS,
   skin: 'dark',
   controls: false,
-  controlBar: {
-    pictureInPictureToggle: false,
-    chaptersButton: false
-  },
+  chaptersButton: false,
+  pictureInPictureToggle: false,
```
ehab-cl, line 13: these are no longer from videojs perspective part of the controlBar?

src/video-player.js
```diff
@@ -368,6 +372,12 @@ class VideoPlayer extends Utils.mixin(Eventable) {
     }
   }
 
+  _initPictureInPicture() {
+    if (!this.playerOptions.pictureInPictureToggle && this.videojs.controlBar) {
+      this.videojs.controlBar.removeChild('pictureInPictureToggle');
```
ehab-cl, line 377: just making sure i understand, we have to do this because the default for videojs is true right?

ehab-cl (approved):
💪

src/config/defaults.js
tsi: they are the same buttons, children of the controlBar, only that we expose the setting as part of our root config

PR #528, opened 2024-01-30
ME-15270-raw-urls-cors-issue

tsi:


src/plugins/cloudinary/models/video-source/video-source.js
```diff
@@ -187,7 +187,7 @@ class VideoSource extends BaseSource {
 
   generateSources() {
     if (this.isRawUrl) {
-      const type = this.sourceTypes().length > 1 ? null : this.sourceTypes()[0];
```
tsi, line 190: This should return `null` for raw-urls, but our new default `sourceTypes` is `['auto']` so this returns `this.sourceTypes()[0]` which causes the issue.
This change fixes that, but breaks a test that expects
```http://res.cloudinary.com/demo/video/upload/vc_vp9/sea_turtle.webm```
to be returned when just `video_codec: 'vp9'` transformation is set
and we are now (since moving to `f_auto` by default) returning
```https://res.cloudinary.com/demo/video/upload/vc_vp9/f_auto:video/elephants```

File: src/plugins/cloudinary/models/video-source/video-source.js [hunk 2]
```diff
@@ -187,7 +187,7 @@ class VideoSource extends BaseSource {
...
+      const type = this.sourceTypes()[0] === 'auto' ? null : this.sourceTypes()[0];
```
tsi, line 190: This should return `null` for raw-urls, but our new default `sourceTypes` is `['auto']` so this returns `this.sourceTypes()[0]` which causes the issue.

This change fixes that, but breaks a test that expects
```http://res.cloudinary.com/demo/video/upload/vc_vp9/sea_turtle.webm```
to be returned when just `video_codec: 'vp9'` transformation is set
and we are now (since moving to `f_auto` by default) returning
```https://res.cloudinary.com/demo/video/upload/vc_vp9/f_auto:video/elephants```

File: src/plugins/cloudinary/models/video-source/video-source.const.js [hunk 3]
```diff
@@ -1,20 +1,38 @@
...
+  'mp4',
+  'mpeg',
+  'mpd',
+  'mxf',
+  'ogv',
+  'webm',
+  'wmv'
+]; // https://cloudinary.com/documentation/video_manipulation_and_delivery#supported_video_formats
+
+export const VIDEO_SUFFIX_REMOVAL_PATTERN = RegExp(`\\.(${COMMON_VIDEO_EXTENSIONS.join('|')})$$`);
```
ehab-cl, line 35: why do you have to remove the extension? would not the `f_auto:video` take precedence anyway?

ehab-cl (approved):
💪

src/plugins/cloudinary/models/video-source/video-source.const.js
tsi: we always did that, I just updated the list 
https://github.com/cloudinary/cloudinary-video-player/blob/edge/src/plugins/cloudinary/models/video-source/video-source.js#L31

It makes sense, in many cases we need the public ID without the extension, i.e. to get a poster image, what I don't get is why we completely ignore it. I mean, if someone did something like 

```
player.source('elephants.webm');
```

IMO we should treat it as if they did `sourceTypes: ['webm']`

ehab-cl: ok now i understand that this is just for getting the public id and not for logic related to formats 👍

PR #535, opened 2024-02-05
ME-14380-lazy-ads

tsi:


docs/vast-vpaid.html
```diff
@@ -67,7 +67,6 @@ <h3 class="mb-4">VAST and VPAID</h3>
       playsinline
       controls
       muted
-      autoplay
```
ehab-cl, line 70: Since the non autoplay was causing a bug it seems that autoplay can be a factor, i would provide two examples one with autoplay and one without.

src/plugins/ima/ima.js
```diff
@@ -1,1775 +1,3 @@
...
-      player.on('volumechange', onVolumeChange_);
-    });
-    this.proxyClickEvents();
-  };
-
-  videojs.registerPlugin('ima', init);
-});
+import 'videojs-contrib-ads';
+import 'videojs-ima';
+import 'videojs-ima/dist/videojs.ima.scss';
```
ehab-cl, line 3: This versus having the library code getting copied is 🕺 🚀

src/plugins/ima/index.js
```diff
@@ -38,11 +36,11 @@ export default function imaPlugin(player, playerOptions) {
   if (Object.keys(playerOptions.ads).length > 0 && typeof player.ima === 'object') {
     if (playerOptions.ads.adsInPlaylist === 'first-video') {
       player.one(PLAYER_EVENT.SOURCE_CHANGED, () => {
-        player.ima.playAd();
+        player.ima.playAdBreak();
```
ehab-cl, line 39: what is the difference between these two methods? just curious or is it just a rename?

ehab-cl (approved):
LGTM 💪

PR #539, opened 2024-02-05
feat: video profiles

jakub-roch:
Video profiles


docs/profiles.html
```diff
@@ -0,0 +1,128 @@
...
+  <h3 class="mb-4">Profiles</h3>
+
+  <p>
+    Profiles are still under development and it's not safe for use on production.<br/>
+    For testing purpose you can use Profiles Feature according to the instructions below:
+    <br/><br/>
+    1. Provide `__profileName` config property using one of default profiles (`loopingVideo`, `videoStream`, `stylish`)<br/>
+    2. You can use your own profile JSON, create folder on your cloud named `profiles` and upload json file
+    - once you have it just provide publicId with extension of your uploaded file as value of `__profileName`
+  </p>
```
tsi, line 58: remove?

src/config/profiles/index.js
```diff
@@ -0,0 +1,9 @@
+import cldDefaultProfile from './cldDefault.json';
+import cldLoopingProfile from './cldLooping.json';
+import cldAdaptiveStreamProfile from './cldAdaptiveStream.json';
+
+export const defaultProfiles = {
+  cldDefault: cldDefaultProfile,
+  cldLooping: cldLoopingProfile,
+  cldAdaptiveStream: cldAdaptiveStreamProfile
+};
```
tsi, line 9:
```suggestion
import cldDefault from './cldDefault.json';
import cldLooping from './cldLooping.json';
import cldAdaptiveStream from './cldAdaptiveStream.json';

export const defaultProfiles = {
  cldDefault,
  cldLooping,
  cldAdaptiveStream
};
```

tsi (approved):
That's much better, looks great! 🚀 

One last thought - we should probably expose this as a separate export as well, see https://github.com/cloudinary/cloudinary-video-player/blob/edge/src/index.es.js
We should do the same for `videoPlayers` as well

PR #541, opened 2024-02-05
fix: profiles error

jakub-roch:
* async module - get named export
* DO NOT return videoPlayer instance if error happened - throw error

### DIFF
--- a/src/video-player-profile.js
+++ b/src/video-player-profile.js
@@ -1,7 +1,7 @@
 import VideoPlayer from './video-player';
 
 export const getProfile = async (cloudName, profile) => {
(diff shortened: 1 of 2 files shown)


ehab-cl (approved):
💪

src/video-player-profile.js
```diff
@@ -30,7 +30,7 @@ const videoPlayerProfile = async (elem, initOptions, ready) => {
   } catch (e) {
     const videoPlayer = new VideoPlayer(elem, initOptions);
     videoPlayer.videojs.error('Invalid profile');
-    return videoPlayer;
+    throw e;
```
ehab-cl, line 33: We need to confirm this behaviour with Raz.

PR #547, opened 2024-02-07
ME-15408-netlify-previews

tsi:


docs/scripts.js
```diff
@@ -31,14 +31,25 @@ var isLocal = window.location.hostname === 'localhost' || window.location.hostna
...
+var isNetlify = window.location.hostname.includes('cld-video-player.netlify.app');
+
 var cdnPrefix = function (source, ver) {
-  var external = source.startsWith('http');
+  var external = source && source.startsWith('http');
+  var previewBuild = ver && ver.startsWith('http');
+
+  if (previewBuild) {
+    return ver;
+  }
```
ehab-cl, line 42: not sure if infra will support this in this project, but basically you can remove the previewBuild function and just do 
```suggestion
  if (ver?.startsWith?.('http')) {
    return ver;
  }
```

Also i would rename ver to version

ehab-cl (approved):
💪

PR #550, opened 2024-02-11
ME-15490-playlist-by-tag

tsi:
This PR fixes an issue that seem to be introduced in 1.10.2 when we introduced `f_auto` as default
https://cloudinary.github.io/cloudinary-video-player/playlist-by-tag-captions.html?ver=latest&min=true


ehab-cl (approved):
💪

PR #551, opened 2024-02-11
ME-15396-chapters-in-Safari

tsi:
This PR fixes a similar issue to the one we saw in the paced-subtitles feature.
When programmatically creating text-tracks, Safari needs to wait a cycle for the track to be ready for cues


ehab-cl (approved):
💪

PR #563, opened 2024-02-22
feat: kareoke style subtitles

tsi:


src/plugins/paced-transcript/index.js
```diff
@@ -46,26 +50,57 @@ function pacedTranscript(config) {
...
+            .slice(i, i + maxWords)
+            .map(word => word.word)
+            .join(' ');
+
+          captions.push({
+            startTime: startTime,
+            endTime: endTime,
+            text: captionText
+          });
+        }
```
ehab-cl, line 103: I understand this problem is inherently imperative and i do not have a specific suggestion here, but see if you can simplify this function. It seems to me that same parts look like they can be reused.

ehab-cl (approved):
LGTM 💪

src/plugins/paced-transcript/index.js
```diff
@@ -13,9 +13,13 @@ function pacedTranscript(config) {
       source.publicId(),
       extendCloudinaryConfig(player.cloudinary.cloudinaryConfig(), { resource_type: 'raw' }),
     ) + '.transcript',
-    maxWords: config.maxWords || 5 // Number of words per caption
+    maxWords: config.maxWords || 5, // Number of words per caption
```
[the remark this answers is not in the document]
tsi: No that's actually a good comment, we shouldn't have any default value set here :)

On the other hand, you'll still see some defaults being set here, i.e. `label`, you can look at it as plugin defaults - because from player perspective there is no default caption or caption label, but from the plugin perspective, when it is enabled, you do need to always have a label, so we can maybe write it more elegantly, but it will still need to be there somehow.

src/plugins/paced-transcript/index.js
tsi: I agree, and adding the `timeOffset` makes this even more noticeable. re-factored

PR #569, opened 2024-03-04
fix: allowUsageReport

tsi:
This PR fixes the `allowUsageReport` configuration param to function as documented


jakub-roch (approved):
Looks good 👍

PR #572, opened 2024-03-05
fix: mimetypes

tsi:
This PR fixes our mimetypes testing logic.
Until now we were simply using the file extension for non ABR, which was ok for some types but bad for others.
This is now aligned with [videojs mimetypes](https://docs.videojs.com/utils_mimetypes.js.html)

### DIFF
--- a/src/plugins/cloudinary/models/video-source/video-source.js
+++ b/src/plugins/cloudinary/models/video-source/video-source.js
@@ -186,8 +186,7 @@ class VideoSource extends BaseSource {
 
   generateSources() {
     if (this.isRawUrl) {
-      const type = this.sourceTypes()[0] === 'auto' ? null : this.sourceTypes()[0];
-      return [this.generateRawSource(this.publicId(), type)];
+      return [this.generateRawSource(this.publicId())];
     }
 
     const srcs = this.sourceTypes().map(sourceType => {
@@ -243,11 +242,11 @@ class VideoSource extends BaseSource {
     return srcs;
   }
 
-  generateRawSource(url, type) {
-    const t = type || url.split('.').pop();
-    const isAdaptive = !!CONTAINER_MIME_TYPES[t];
-    if (isAdaptive) {
-      type = CONTAINER_MIME_TYPES[t][0];
+  generateRawSource(url) {
+    let type = url.split('.').pop();
+    const isAdaptive = ['mpd', 'm3u8'].indexOf(type) !== -1;
+    if (CONTAINER_MIME_TYPES[type]) {
+      type = CONTAINER_MIME_TYPES[type];
     } else {
       type = type ? `video/${type}` : null;
     }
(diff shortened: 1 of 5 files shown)


src/plugins/cloudinary/models/video-source/video-source.js
```diff
@@ -244,10 +244,12 @@ class VideoSource extends BaseSource {
   }
 
   generateRawSource(url, type) {
-    const t = type || url.split('.').pop();
-    const isAdaptive = !!CONTAINER_MIME_TYPES[t];
-    if (isAdaptive) {
-      type = CONTAINER_MIME_TYPES[t][0];
+    type = type || url.split('.').pop();
+
+    const isAdaptive = ['hls', 'dash', 'mpd', 'm3u8'].indexOf(type) !== -1;
```
jakub-roch, line 249: what do you say about moving this list to video source const?

`export const ADAPTIVE_SOURCES = ['hls', 'dash', 'mpd', 'm3u8'];`

src/utils/cloudinary.js
```diff
@@ -34,7 +34,7 @@ const setError = (that, res) => {
 };
 
 const setVideoSrc = (that, srcs) => {
-  console.log('Trying urls: ' + JSON.stringify(srcs));
+  console.log('Trying sources: ', srcs);
```
jakub-roch, line 37: I see it was previously there but can we get rid of it or hide it behind some debug mode?

jakub-roch (approved):
Looks good but please see my comments

src/plugins/cloudinary/models/video-source/video-source.utils.js
```diff
@@ -6,16 +6,12 @@ import { some } from '../../../../utils/array';
...
 export function formatToMimeTypeAndTransformation(format) {
   const [container, codec] = format.toLowerCase().split('\/');
-  let result = CONTAINER_MIME_TYPES[container];
-  let transformation = null;
-
-  if (!result) {
-    result = [`video/${container}`, transformation];
-  }
+  const mimetype = CONTAINER_MIME_TYPES[container] || `video/${container}`;
+  let result = [mimetype];
```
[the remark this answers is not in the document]
tsi: I really prefer avoiding multiple returns where possible

src/plugins/cloudinary/models/video-source/video-source.utils.js
```diff
@@ -6,16 +6,12 @@ import { some } from '../../../../utils/array';
...
 
   if (codec) {
-    transformation = codecToSrcTransformation(codec);
-    result = [`${result[0]}; codecs="${codecShorthandTrans(codec)}"`, transformation];
+    const transformation = codecToSrcTransformation(codec);
+    result = [`${CONTAINER_MIME_TYPES[container]}; codecs="${codecShorthandTrans(codec)}"`, transformation];
```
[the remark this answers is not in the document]
tsi: That's not clear, formatting was messed up?
Anyway,  `container` should never be undefined, as it iterates over sourceTypes array

src/utils/cloudinary.js
tsi: I'd like to do that, but we don't have dev-mode yet (I plan to add it) and this is logged only on error and is now much less noisy and helps debugging, so I guess it's ok for now

PR #573, opened 2024-03-06
fix: poster image for raw URLs

tsi:


src/plugins/cloudinary/index.js
```diff
@@ -255,9 +255,10 @@ class CloudinaryContext {
 
     const refresh = () => {
       const src = this.source();
-      const { posterColor } = this.player.cloudinary.posterOptions();
-      if (posterColor) {
-        this.disablePoster(posterColor);
+      const posterOptions = Object.assign({}, this.player.cloudinary.posterOptions(), src.getInitOptions().poster);
+
+      if (posterOptions.posterColor) {
+        this.disablePoster(posterOptions.posterColor);
```
ehab-cl, line 261: I am not sure i understand the logic here even before your change... like if posterColor is supplied as an option then we continue to disable it, why is that?

ehab-cl (approved):

src/plugins/cloudinary/index.js
tsi: instead of `disablePoster` you can call it `resetPosterAndSetPosterColorIfConfigured` :) 
I really don't like the whole poster flow, but like you said yesterday, we shouldn't/can't fix everything now

PR #576, opened 2024-03-12
fix: styled subtitles browser compatibility

tsi:
Managed to support most styling properties without disabling nativeTextTracks, so it’s also supported in iOS full screen

What’s supported everywhere - basic low level text styling - i.e. font-family, font-size, color, text-shadow etc, also for highlighted word.

Partially supported - i.e. in all browsers except Safari with nativeTextTracks - mostly position related property - i.e. gravity, position, margin etc


ehab-cl (approved):
LGTM 💪

PR #588, opened 2024-03-19
perf!: default export to be base videoPlayer

tsi:
BREAKING CHANGE: the default ES import is now only the player core plugins need to be explicitly imported


.npmignore
```diff
@@ -1,11 +0,0 @@
-!dist/
```
ehab-cl, line 1: why did you delete the npmignore file?

ehab-cl (approved):
💪

.npmignore
tsi: I wanted to stop pushing our sourcemap files to npm, so I tried doing it via npmignore, then I realized it was redundant ever since we're using the `files` field in package.json
Also [this](https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d) :)

PR #590, opened 2024-03-20
perf: cleanup codec logic

tsi:
* Cleanup codec logic
* Update to public `videojs-contrib-dash`, since it is maintained again.


src/plugins/cloudinary/index.js
```diff
@@ -86,8 +83,8 @@ class CloudinaryContext {
       _source = src;
 
       const isDashRequired = options.sourceTypes && options.sourceTypes.some(s => s.includes('dash'));
-      if (plugins.dashPlugin && isDashRequired) {
-        plugins.dashPlugin().then(() => {
+      if (isDashRequired) {
```
ehab-cl, line 86: 🕺

src/plugins/cloudinary/models/video-source/video-source.utils.js
```diff
@@ -1,36 +1,38 @@
...
   if (!res) {
-    res = format.split('\/').shift();
+    res = format.split('/').shift();
   }
 
   return res;
 }
 
-const isContainCodec = (value) => value && some(Object.values(VIDEO_CODEC), (codec) => value.includes(codec));
+const isContainCodec = value =>
```
ehab-cl, line 31:
```suggestion
const doesContainCodec = value =>
```

ehab-cl: or just `hasCodec` maybe

src/plugins/cloudinary/models/video-source/video-source.utils.js
```diff
@@ -1,36 +1,38 @@
...
+  value && some(Object.values(VIDEO_CODEC), codec => value.includes(codec));
 
-const hasCodecSrcTrans = (transformations) => some(['video_codec', 'streaming_profile'], (key) => isKeyInTransformation(transformations, key));
+const hasCodecSrcTrans = transformations =>
+  some(['video_codec', 'streaming_profile'], key => isKeyInTransformation(transformations, key));
```
ehab-cl, line 35: For later or maybe even now you can just refactor this usage of some and use the es6 built in some method for arrays.

ehab-cl (approved):
💪

src/plugins/cloudinary/common.js
```diff
@@ -102,109 +102,13 @@ const cloudinaryErrorsConverter = ({ errorMsg, publicId, cloudName, statusCode }
   return error;
 };
 
-const codecShorthandTrans = short => {
```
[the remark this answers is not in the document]
tsi: We decided not to test for specific codec support anymore, since almost all browsers now support them all.
With the default `f_auto` there shouldn't be an issue ever.
There are some edge cases in which old Safari's for example might try playing an AV1 and will have to fall-back to the next `sourceType`, which will cause a console error but nothing fatal. And if that's a concern we leace it to the implementor to prevent.

PR #595, opened 2024-03-26
perf: utils refactor & cleanup

tsi:


src/plugins/playlist/utils/dom.js
```diff
@@ -0,0 +1,8 @@
+const wrap = (el, wrapper) => {
+  el.parentNode.insertBefore(wrapper, el);
+  wrapper.appendChild(el);
```
ehab-cl, line 3: I know that this is an old function: but does not the code eventually leaves duplicate el one that is wrapped and one that is not wrapped? I mean this seems intended but i find it a bit weird, i thought this function would be deleting the "unwrapped element"?

src/plugins/playlist/utils/time.js
```diff
@@ -0,0 +1,39 @@
...
+
+    timestamp = Date.UTC(struct[1], struct[2], struct[3], struct[4], struct[5] + minutesOffset, struct[6], struct[7]);
+  } else {
+    timestamp = NaN;
+  }
+
+  return timestamp;
+};
+
+export { parseISO8601 };
```
ehab-cl, line 39: Oh later we should definitely use a parser form a library 😄

src/utils/cloudinary.js
```diff
@@ -74,38 +73,6 @@ const handleCldError = (that, options) => {
   }
 };
 
-// for IE 11
-function parseUri(str) {
```
ehab-cl, line 78: 🕺

ehab-cl (approved):
💪

src/plugins/playlist/utils/dom.js
tsi: No, it inserts `wrapper` before `el` and then moves `el` into `wrapper`

src/plugins/playlist/utils/time.js
tsi: Well, yeah, there's so much more we can do... we can go forever

src/plugins/playlist/utils/dom.js
ehab-cl: @tsi but it does not remove el from its parent right?

PR #603, opened 2024-04-07
build: add default profiles json files to build output in es version

jakub-roch:
Export default profiles as raw json files so they can be imported directly in BFF.

image

### DIFF
--- a/webpack/es6.config.js
+++ b/webpack/es6.config.js
@@ -31,7 +31,7 @@ module.exports = merge(webpackCommon, {
     new CopyWebpackPlugin({
       patterns: [{
         from: path.resolve(__dirname, '../src/config/profiles'),
-        to: `${outputPath}/default-player-profiles`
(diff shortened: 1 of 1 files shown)


webpack/es6.config.js
```diff
@@ -16,14 +19,23 @@ module.exports = merge(webpackCommon, {
...
     library: {
       type: 'module'
     }
   },
 
+  plugins: [
+    new CopyWebpackPlugin({
+      patterns: [{
+        from: path.resolve(__dirname, '../src/config/profiles'),
+        to: `${outputPath}/default-player-profiles`
```
tsi, line 34:
```suggestion
        to: `${outputPath}/profiles`
```

more than enough 😄

tsi (approved):
👍

webpack/es6.config.js
jakub-roch: 👍

PR #604, opened 2024-04-07
fix: simplify eslint config (ME-15921)

tsi:


.eslintrc.js
```diff
@@ -1,275 +1,29 @@
...
-      "error",
-      "never"
-    ]
+  rules: {
+    indent: ['error', 2, { SwitchCase: 1 }],
+    'linebreak-style': ['error', 'unix'],
+    quotes: ['error', 'single'],
+    semi: ['warn', 'always'],
+    'no-console': 'off',
+    'no-unused-vars': 'warn',
```
ehab-cl, line 26: Do not you want to error on unused vars?

.eslintrc.js
```diff
@@ -1,275 +1,29 @@
...
-    "yoda": [
-      "error",
-      "never"
-    ]
+  rules: {
+    indent: ['error', 2, { SwitchCase: 1 }],
+    'linebreak-style': ['error', 'unix'],
+    quotes: ['error', 'single'],
+    semi: ['warn', 'always'],
+    'no-console': 'off',
```
ehab-cl, line 25: I would at least leave a warning here.

ehab-cl (approved):

.eslintrc.js
tsi: I was trying to simplify without causing any change, that's the way it was before.
We can open this for discussion, but changes here will require adjustments to the code

.eslintrc.js
tsi: same ⬆️

.eslintrc.js
ehab-cl: I agree it is a different scope, I think that in general you want to forbid the usage of console.log to ensure that debugging statements are not left by mistake, if you agree with me, you can just create a task for later and put at the top of the backlog.

.eslintrc.js
ehab-cl: Same answer for the next comment, different scope, if you agree please create a task if not then lets discuss it later.

.eslintrc.js
tsi: I completely agree, in fact my current task is wrapping console.logs with a debug-mode condition :) 
But opening a discussion around such decisions in the video-player context is going to be a huge overhead. I'd say we should have a company level eslint standard (if we don't already) and discuss it at that level.

PR #627, opened 2024-05-01
fix: split chunks styles

tsi:


package.json
```diff
@@ -72,7 +72,7 @@
       },
       {
         "path": "./lib/all.js",
-        "maxSize": "280kb"
+        "maxSize": "290kb"
```
ehab-cl, line 75: has the bundle size eventually increased by 10 kb? I am assuming this is to the all bundle size?

ehab-cl (approved):
Please mark documentation in the relevant ticket and explain what in the ticket how to the documentation needs to adapt.

package.json
tsi: yes, only for `/all` because it contains chunks that now contain some "inline" styles. the core package didn't change

PR #637, opened 2024-05-28
fix: bump video analytics, trigger analytics event only once

jakub-roch:
Bump video analytics package version
Trigger source change event only once


tsi (approved):
👍

PR #639, opened 2024-06-04
fix: use videojs events & re-trigger to analytics package with custom…

jakub-roch:
Unfortunately video.js intercept all video events on HTML tag which causes problems only on iOS mobile browsers.
The only solution that works is to trigger our own events (custom events) from VP side.

### DIFF
--- a/src/plugins/cloudinary-analytics/index.js
+++ b/src/plugins/cloudinary-analytics/index.js
@@ -1,5 +1,5 @@
 import videojs from 'video.js';
-import { connectCloudinaryAnalytics } from '../../../../cloudinary-video-analytics';
+import { connectCloudinaryAnalytics } from 'cloudinary-video-analytics';
 import { PLAYER_EVENT } from '../../utils/consts';
 
 class CloudinaryAnalytics {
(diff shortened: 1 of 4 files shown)


src/plugins/cloudinary-analytics/index.js
```diff
@@ -1,10 +1,14 @@
...
 import { PLAYER_EVENT } from '../../utils/consts';
 
 class CloudinaryAnalytics {
   constructor(player) {
     this.player = player;
-    this.cloudinaryAnalytics = connectCloudinaryAnalytics(this.player.videoElement);
+    this.shouldUseCustomEvents = videojs.browser.IS_IOS;
+    this.cloudinaryAnalytics = connectCloudinaryAnalytics(this.player.videoElement, {
+      customEvents: this.shouldUseCustomEvents,
+    });
```
tsi, line 11: @jakub-roch [this ticket](https://github.com/videojs/video.js/issues/6679) suggests that you should listen to events on the `player` and not on the `player.videoElement`, did you try that maybe?

tsi (approved):

tsi (approved):

PR #662, opened 2024-06-17
feat: support profiles in default videoPlayer method

tsi:


tsi:
Usage example: https://codepen.io/tsi/pen/xxNpmNY?editors=0010

PR #663, opened 2024-06-24
fix: separate edge/master changelogs

tsi:
Have 2 separate changelogs will make merging much simpler, for now I'm keeping the 2 branches flow


ehab-cl (approved):
💪

PR #671, opened 2024-07-02
fix: chapters innerHTML to innerText

tsi:
This PR switches the chapters plugin from using innerHTML to innerText for safer interpolation of user provided content.


ehab-cl (approved):
💪

PR #673, opened 2024-07-03
fix: custom profile docs example

jakub-roch:


tsi (approved):
LGTM 💪 

@jakub-roch could we use a profile that is more “noticeable”? maybe something like [we have here](https://codepen.io/tsi/pen/xxNpmNY)?

PR #678, opened 2024-07-14
feat: new 'player' method with support async loading & profiles

jakub-roch:
New `player` async method with support of profiles feature.
Old method `videoPlayerWithProfiles` is marked as deprecated and should be removed soon.
Important: docs will be updated once we have URL template for custom profiles.


src/index.js
```diff
@@ -2,6 +2,7 @@ import 'assets/styles/main.scss';
 import pick from 'lodash/pick';
 import VideoPlayer from './video-player';
 import createVideoPlayerProfile from './video-player-profile';
+import createPlayer from './player';
```
tsi, line 5: Why are we having two separate methods that does exactly the same?
If we really must, just do `export const videoPlayerWithProfile =  player`

src/index.js
```diff
@@ -22,10 +23,14 @@ const getVideoPlayers = config => (selector, playerOptions, ready) =>
 
 const getVideoPlayerWithProfile = config => (id, playerOptions, ready) => createVideoPlayerProfile(id, getConfig(playerOptions, config), ready);
 
+const getPlayer = config => (id, playerOptions, ready) => createPlayer(id, getConfig(playerOptions, config), ready);
+
 export const videoPlayer = getVideoPlayer();
 export const videoPlayers = getVideoPlayers();
 export const videoPlayerWithProfile = getVideoPlayerWithProfile();
 
+export const player = getPlayer();
```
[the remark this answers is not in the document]
tsi: ➕

src/player.js
```diff
@@ -0,0 +1,34 @@
+import VideoPlayer from './video-player';
+import { defaultProfiles } from './config/profiles';
+
+export const getProfile = async (cloudName, profile) => {
+  if (Object.keys(defaultProfiles).includes(profile)) {
+    return defaultProfiles[profile];
+  }
+
+  console.warn('Custom profiles loading mechanism will be changed soon');
```
tsi, line 9: ?

As a developer, if I see this, what do I do?

src/video-player-profile.js
```diff
@@ -10,6 +10,8 @@ export const getProfile = async (cloudName, profile) => {
 };
 
 const videoPlayerProfile = async (elem, initOptions, ready) => {
+  console.warn('videoPlayerProfile method is DEPRECATED and will be removed soon, please use new `player` method instead');
```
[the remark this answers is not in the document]
ehab-cl: I thought this was the standard regular method, please ignore my comment.

src/index.js
jakub-roch: there is small difference between them (profile property as optional one) and there is possibility that some people are using it - in this case first we should deprecate it and later after some time just remove it

tsi (approved):
✅

PR #684, opened 2024-07-17
vp test: adding new test to handle ESM page + small refactoring

ShayLevi:
Adding new test to handle ESM page example. We're checking that no errors are thrown to console (besides the pages which we expect to get ones).
Also contains small refactoring - moving common functions from tests to helpers.


tsi (approved):
👍

refael-m (approved):

PR #689, opened 2024-08-07
fix: docs for profiles & analytics, analytics options

jakub-roch:
Fix: docs for profiles & analytics
analytics options


src/plugins/cloudinary-analytics/index.js
```diff
@@ -26,9 +26,9 @@ class CloudinaryAnalytics {
     if (metadata.cloudName && metadata.publicId) {
       this.currentVideMetadata = metadata;
       this.cloudinaryAnalytics.startManualTracking(metadata, {
+        ...(typeof this.analyticsOptions === 'object' ? this.analyticsOptions : {}),
```
jakub-roch, line 29: it doesn't need to be checked against null as if it will be null if condition won't allow to turn on analytics

tsi (approved):
LGTM ✅

PR #696, opened 2024-10-01
feat: add url template for video player profiles

jakub-roch:
* Add url template for video player profiles with support for staging env
* Remove deprecated method `videoPlayerWithProfile`

### DIFF
--- a/docs/profiles.html
+++ b/docs/profiles.html
@@ -38,7 +38,7 @@
         profile: 'myCustomProfile',
       });
 
-      playerWithCustomProfile.source('sea_turtle');
+      playerWithCustomProfile.source('samples/cld-sample-video');
     }, false);
   </script>
 </head>
@@ -121,7 +121,7 @@
             profile: 'myCustomProfile',
           });
 
-          playerWithCustomProfile.source('sea_turtle');
+          playerWithCustomProfile.source('samples/cld-sample-video');
         }, false);
       </code>
     </pre>
--- a/docs/profiles.html
+++ b/docs/profiles.html
@@ -31,6 +31,15 @@
 
       playerWithDefaultProfile.source('sea_turtle');
     }, false);
+
+    window.addEventListener('load', async function() {
+      const playerWithCustomProfile = await cloudinary.player('player-custom-profile', {
+        cloudName: 'demo',
+        profile: 'myCustomProfile',
+      });
+
+      playerWithCustomProfile.source('sea_turtle');
+    }, false);
   </script>
 </head>
 <body>
@@ -75,6 +84,44 @@
           });
 
           playerWithDefaultProfile.source('sea_turtle');
+        }, false);
+      </code>
+    </pre>
+
+  <h5>Player with custom profile</h5>
+  <video
+    id="player-custom-profile"
+    playsinline
+    controls
+    autoplay
+    muted
+    class="cld-video-player"
+    width="500">
+  </video>
+
+  <h3 class="mt-4">Example Code:</h3>
+
+  <pre>
+    <code class="language-html">
+
+      &lt;video
+        id="player-custom-profile"
+        controls
+        autoplay
+        muted
+        class="cld-video-player"
+        width="500"&gt;
+      &lt;/video&gt;
+
+      </code>
+      <code class="language-javascript">
+        window.addEventListener('load', async function() {
+          const playerWithCustomProfile = await cloudinary.player('player-custom-profile', {
+            cloudName: 'demo',
+            profile: 'myCustomProfile',
+          });
+
+          playerWithCustomProfile.source('sea_turtle');
         }, false);
       </code>
     </pre>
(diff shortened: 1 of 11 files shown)


src/player.js
```diff
@@ -2,22 +2,21 @@ import VideoPlayer from './video-player';
...
   }
 
-  if (isRawUrl(profile)) {
-    return await fetch(profile, { method: 'GET' }).then(res => res.json());
-  }
-
-  throw new Error('Custom profiles will be supported soon, please use one of default profiles: "cldDefault", "cldLooping" or "cldAdaptiveStream"');
+  const cloudinaryDomain = secureDistribution === 'res-staging.cloudinary.com' ? secureDistribution : 'res.cloudinary.com';
+  const profileUrl = isRawUrl(profile) ? profile :
+    `https://${cloudinaryDomain}/${cloudName}/_applet_/video_service/video_player_profiles/${profile}.json`;
```
tsi, line 12: There are many considerations to take when building Cloudinary URLs (`private_cdn`, `cname` etc), you should use the `url-gen` SDK to handle that, we already import `createCloudinaryLegacyURL` for video URLs

https://github.com/cloudinary/js-url-gen/blob/master/src/backwards/createCloudinaryLegacyURL.ts

tsi (changes requested):

docs/profiles.html
```diff
@@ -78,6 +87,44 @@ <h3 class="mt-4">Example Code:</h3>
...
+            profile: 'myCustomProfile',
+          });
+
+          playerWithCustomProfile.source('samples/cld-sample-video');
+        }, false);
+      </code>
+    </pre>
   </div>
 
 </body>
```
ehab-cl, line 130: I suggest adding another example with a custom profile and overriding configuration code to make sure it works as expected (It could be same page but two different players and i think this is even more favourable)

src/player.js
```diff
@@ -1,23 +1,32 @@
...
   if (Object.keys(defaultProfiles).includes(profile)) {
     return defaultProfiles[profile];
   }
 
-  if (isRawUrl(profile)) {
-    return await fetch(profile, { method: 'GET' }).then(res => res.json());
-  }
+  const urlPrefix = unsigned_url_prefix(
+    null,
+    initOptions.cloudName || initOptions.cloud_name,
```
ehab-cl, line 13:
```suggestion
    initOptions.cloudName ?? initOptions.cloud_name,
```

src/player.js
```diff
@@ -1,23 +1,32 @@
...
+    initOptions.cdn_subdomain,
+    initOptions.secure_cdn_subdomain,
+    initOptions.cname,
+    initOptions.secure,
+    initOptions.secure_distribution,
+  );
 
-  throw new Error('Custom profiles will be supported soon, please use one of default profiles: "cldDefault", "cldLooping" or "cldAdaptiveStream"');
+  const profileUrl = isRawUrl(profile) ? profile : `${urlPrefix}/_applet_/video_service/video_player_profiles/${profile}.json`;
+  return await fetch(profileUrl, { method: 'GET' }).then(res => res.json());
```
ehab-cl, line 23: I think the await here and the async in the function declaration are both redundant, but especially here it is really not needed
```suggestion
  return fetch(profileUrl, { method: 'GET' }).then(res => res.json());
```

ehab-cl (approved):
LGTM 💪

src/player.js
jakub-roch: changed, I would like to keep entire method as async just in case - so w need to have async declaration on the top but there await can be removed

src/player.js
jakub-roch: 👍

docs/profiles.html
jakub-roch: good point, added another example with overrides

src/player.js
```diff
@@ -1,23 +1,32 @@
...
+  const urlPrefix = unsigned_url_prefix(
+    null,
+    initOptions.cloudName ?? initOptions.cloud_name,
+    initOptions.private_cdn,
+    initOptions.cdn_subdomain,
+    initOptions.secure_cdn_subdomain,
+    initOptions.cname,
+    initOptions.secure,
+    initOptions.secure_distribution,
+  );
```
tsi, line 20: @jakub-roch instead of creating that list you should use `CLOUDINARY_CONFIG_PARAM`, [like in getConfig()](https://github.com/cloudinary/cloudinary-video-player/blob/edge/src/index.js#L8)

tsi (approved):

src/player.js
jakub-roch: image
this helper requires specific properties, some of them are ignored like ```'shorten',
  'sign_url',
  'api_secret',
  'url_suffix',
  'use_root_path',
  'auth_token'```

as I can see `cloudinaryConfig` provides same values as from initOptions

PR #699, opened 2024-10-14
feat: add internal analytics about new method & profiles

jakub-roch:
Add internal analytics about usage of new method & profiles feature

### DIFF
--- a/src/video-player.js
+++ b/src/video-player.js
@@ -55,7 +55,6 @@ class VideoPlayer extends Utils.mixin(Eventable) {
     this.videoElement = getResolveVideoElement(elem);
 
     this.options = extractOptions(this.videoElement, initOptions);
-    console.log('1111', this.options);
 
     this._videojsOptions = this.options.videojsOptions;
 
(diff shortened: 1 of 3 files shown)


src/video-player.js
```diff
@@ -55,6 +55,7 @@ class VideoPlayer extends Utils.mixin(Eventable) {
     this.videoElement = getResolveVideoElement(elem);
 
     this.options = extractOptions(this.videoElement, initOptions);
+    console.log('1111', this.options);
```
ehab-cl, line 58:
```suggestion
```

src/video-player.js
```diff
@@ -112,6 +113,7 @@ class VideoPlayer extends Utils.mixin(Eventable) {
       return;
     }
     try {
+      const internalAnalyticsMetadata = this.options.videojsOptions._internalAnalyticsMetadata || {};
```
ehab-cl, line 116:
```suggestion
      const internalAnalyticsMetadata = this.options.videojsOptions._internalAnalyticsMetadata ?? {};
```

ehab-cl: does not also make sense that something that is Cloudinary related is stored under `videojsOptions` ?

ehab-cl (approved):

src/video-player.js
jakub-roch: 👍

src/video-player.js
jakub-roch: good point, I moved it to playerOptions as its internal property not from videojs

PR #737, opened 2024-11-25
feat: allow transcript url

tsi:
This Pr adds the ability to explicitly configure transcript source URLs

i.e.
```
captions: {
  label: "Lincoln's Transcript",
  url: 'https://res.cloudinary.com/demo/raw/upload/lincoln.transcript',
  wordHighlight: true,
  maxWords: 8,
}
```

### DIFF
--- a/src/utils/cloudinary.js
+++ b/src/utils/cloudinary.js
@@ -101,13 +101,13 @@ const isKeyInTransformation = (transformation, key) => {
 
 const addTextTracks = (tracks, videojs) => {
   tracks.forEach(track => {
-    if (track.src) {
+    if (track.src && track.src.endsWith('.vtt')) {
       fetch(track.src, GET_ERROR_DEFAULT_REQUEST).then(r => {
         if (r.status >= 200 && r.status <= 399) {
           videojs.addRemoteTextTrack(track, true);
         }
       });
-    } else if (videojs.pacedTranscript) {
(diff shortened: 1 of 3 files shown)


src/utils/cloudinary.js
```diff
@@ -101,13 +101,13 @@ const isKeyInTransformation = (transformation, key) => {
...
   tracks.forEach(track => {
-    if (track.src) {
+    if (track.src && track.src.endsWith('.vtt')) {
       fetch(track.src, GET_ERROR_DEFAULT_REQUEST).then(r => {
         if (r.status >= 200 && r.status <= 399) {
           videojs.addRemoteTextTrack(track, true);
         }
       });
-    } else if (videojs.pacedTranscript) {
+    } else if (videojs.pacedTranscript && (!track.src || track.src.endsWith('.transcript'))) {
```
ehab-cl, line 110: how does it work? i do not see any mention of url property is it being translated somewhere to src?

ehab-cl (approved):
💪

src/utils/cloudinary.js
tsi: yes, our original API was using `url` for VTTs, way before we added transcript support
https://github.com/cloudinary/cloudinary-video-player/blob/edge/src/video-player.js#L225

PR #741, opened 2024-11-26
feat: auto-fetch transcripts from language

tsi:
This PR adds the ability to auto-fetch transcripts based on language codes

### DIFF
--- a/src/plugins/paced-transcript/index.js
+++ b/src/plugins/paced-transcript/index.js
@@ -3,16 +3,18 @@ import { getCloudinaryUrl, extendCloudinaryConfig } from 'plugins/cloudinary/com
 function pacedTranscript(config) {
   const player = this;
   const source = player.cloudinary.source();
+  const baseUrl = getCloudinaryUrl(
+    source.publicId(),
+    extendCloudinaryConfig(player.cloudinary.cloudinaryConfig(), { resource_type: 'raw' })
+  );
 
   const options = {
     kind: config.kind || 'captions',
     label: config.label || 'Captions',
     default: config.default,
     srclang: config.srclang || 'en',
-    src: config.src || getCloudinaryUrl(
-      source.publicId(),
-      extendCloudinaryConfig(player.cloudinary.cloudinaryConfig(), { resource_type: 'raw' })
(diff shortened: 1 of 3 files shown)


src/plugins/paced-transcript/index.js
```diff
@@ -23,30 +25,35 @@ function pacedTranscript(config) {
 
   // Load the transcription file
   const initTranscript = async () => {
+    let transcriptResponse;
```
ehab-cl, line 28: Instead of re-assigning i would introduce a fallback fetch function and use it (even if it is just introduced in this file, i think it makes the code much easier to read)

```
const fallbackFetch = async (url, fallbackUrl) => {
  try {
    const response = await fetch(url);
    if (!response.ok) {
      throw new Error(`Failed fetching from ${url} with status code ${response.status}`);
    }
    return response;
  } catch (err) {
    try {
      const fallbackResponse = await fetch(fallbackUrl);
      if (!fallbackResponse.ok) {
        throw new Error(`Failed fetching from ${fallbackUrl} with status code ${fallbackResponse.status}`);
      }
      return fallbackResponse;
    } catch (err) {
      throw new Error(`Failed fetching from  both ${url} and ${fallbackUrl}`);
    }
  }
};


```

You might still want to catch the error when it fails to catch from both urls and either silent it or let it crash the VP?

src/plugins/paced-transcript/index.js
```diff
@@ -23,30 +25,35 @@ function pacedTranscript(config) {
...
-      });
-
+      transcriptResponse = await fetch(options.src);
+      if (!transcriptResponse.ok) {
+        throw new Error(`loading transcription from ${options.src} failed, trying fallback URL`);
+      }
     } catch (error) {
-      console.error('Error loading transcription file:', error);
+      console.error(error);
+      transcriptResponse = await fetch(options.fallbackSrc);
```
ehab-cl, line 36: Please note that you are not catching the error here... i do not know what will happen when it gets propagated .. maybe it crashes the VP if so is that what we want?

ehab-cl (approved):
💪

PR #743, opened 2024-12-02
feat: support srt subtitle format

tsi:
This PR introduces a new `srtTextTracks` plugin that takes an SRT file, fetches it and convert it so it can be fed into `player.addRemoteTextTrack()`.

The code is roughly based on our styled-text-tracks & paced-transcript plugins and the conversion logic is inspired by https://github.com/silviapfeiffer/silviapfeiffer.github.io/blob/master/index.html#L150

### DIFF
--- a/docs/es-modules/subtitles-and-captions.html
+++ b/docs/es-modules/subtitles-and-captions.html
@@ -127,8 +127,8 @@
       import 'cloudinary-video-player/cld-video-player.min.css';
       import 'cloudinary-video-player/playlist';
 
-      const player = cloudinary.videoPlayer('player', {
-        cloud_name: 'prod'
+      const player = videoPlayer('player', {
+        cloudName: 'prod'
       });
 
       player.source(
@@ -213,7 +213,7 @@
       playlist.playlist(playlistSources, playlistOptions);
 
       // Paced
-      const pacedPlayer = cloudinary.videoPlayer('paced', {
+      const pacedPlayer = videoPlayer('paced', {
         cloudName: 'prod'
       });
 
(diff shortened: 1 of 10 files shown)


docs/es-modules/subtitles-and-captions.html
```diff
@@ -116,32 +116,32 @@ <h4 class="mt-4 mb-2">Karaoke player</h4>
       import 'cloudinary-video-player/cld-video-player.min.css';
       import 'cloudinary-video-player/playlist';
 
-      const player = videoPlayer('player', {
-        cloudName: 'demo'
+      const player = cloudinary.videoPlayer('player', {
+        cloud_name: 'prod'
```
jakub-roch, line 120: is it expected change?

tsi: The cloud change?  yes
Adding `cloudinary`? no, but It doesn't really matter

jakub-roch: I meant mostly about changing `cloudName` to `cloud_name`, isnt `cloudName` expected prop?

src/plugins/srt-text-tracks/srt-text-tracks.js
```diff
@@ -0,0 +1,107 @@
...
+    });
+  };
+
+  player.one('loadedmetadata', () => {
+    initSRT();
+  });
+}
+
+// SRT to WebVTT conversion functions
+const srt2webvtt = (data) => {
```
jakub-roch, line 47: what about using some ready lib like - https://www.npmjs.com/package/srt-parser-2 ?

jakub-roch (approved):
Looks good, just please see my comment about lib

src/plugins/srt-text-tracks/srt-text-tracks.js
tsi: Good suggestion 💪

PR #746, opened 2024-12-04
fix: update dependencies

tsi:


ehab-cl (approved):
💪

src/video-player.utils.js
```diff
@@ -41,7 +41,7 @@ export const getResolveVideoElement = (elem) => {
 
     try {
       elem = document.querySelector(`#${cssEscape(id)}`) || videojs.getPlayer(id);
-    } catch (e) {
+    } catch (err) { // eslint-disable-line no-unused-vars
       elem = null;
```
ehab-cl, line 45: I would console.warn the error instead of the eslint disable line ... since anyway we have other console.warn statements .... later we need to start thinking how we should report these errors somewhere

src/utils/attributes-normalizer.js
```diff
@@ -15,7 +15,7 @@ function coerce(string) {
   // Complex object
   try {
     val = JSON.parse(string);
-  } catch (e) {
+  } catch (e) { // eslint-disable-line no-unused-vars
```
ehab-cl, line 18: I would console.warn the error instead of the eslint disable line ... since anyway we have other console.warn statements .... later we need to start thinking how we should report these errors somewhere

tsi: In principle I agree, but it seems like in this case the author used try...catch to silence the failing `JSON.parse()`
I'll try thinking of a better way to write this

src/video-player.utils.js
tsi: see previous

PR #747, opened 2024-12-05
fix: programatic text-tracks in Safari

tsi:
This PR disables `nativeTextTracks` in Safari.
We did everything possible to use the native text tracks, but we now realize it never worked good enough so falling back to our emulated text tracks.


ehab-cl (approved):
💪

PR #765, opened 2024-12-16
feat: light esm build

tsi:


docs/es-modules/all.html
```diff
@@ -0,0 +1,64 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>Cloudinary Video Player</title>
```
ehab-cl, line 5: Not critical but the title should match the example name ....same comment for the second example as well.

docs/es-modules/light.html
```diff
@@ -0,0 +1,65 @@
...
+      <p class="mt-4">
+        <a href="https://cloudinary.com/documentation/cloudinary_video_player#installation"
+          >Full documentation</a
+        >
+      </p>
+    </div>
+
+    <script type="module">
+      import { videoPlayer } from 'cloudinary-video-player/light';
+      import 'cloudinary-video-player/light/chapters';
```
ehab-cl, line 42: would this work `import 'cloudinary-video-player/chapters'`? if so i think its preferable since the current syntax implies that this is a light plugin which is not true in my eyes (The plugin is the same)

Also this way the imports are not consistent with the css import

src/utils/get-analytics-player-options.js
```diff
@@ -20,11 +20,11 @@ const getCloudinaryOptions = (cloudinaryOptions = {}) => ({
 });
 
 const getTranscriptOptions = (textTracks = {}) => {
-  const tracksArr = [textTracks.captions, ...textTracks.subtitles];
+  const tracksArr = [textTracks.captions, ...(textTracks.subtitles || [])];
   return {
     textTracks: hasConfig(textTracks),
     textTracksLength: tracksArr.length,
-    textTracksOptions: hasConfig(textTracks.options) || Object.keys(textTracks.options).join(','),
+    textTracksOptions: hasConfig(textTracks.options) && Object.keys(textTracks.options).join(','),
```
ehab-cl, line 27: how these two changes are related to exporting light version?

webpack/es-light.config.js
```diff
@@ -0,0 +1,17 @@
+const { merge } = require('webpack-merge');
+const webpackConfig = require('./es6.config');
+const path = require('path');
+
+delete webpackConfig.entry; // overwrite
```
ehab-cl, line 5: could be that there is something in webpack-merge api that allows you to overwrite when merging which maybe can be easier to read as it is more explict ? (Just a thought, i did not really go and checked their API)

ehab-cl (approved):
💪

src/utils/get-analytics-player-options.js
tsi: They don't, just fixed a warning I got while working on this

docs/es-modules/light.html
tsi: I agree, unfortunately it won't work since both /chapters and /light/chapters are importing videojs, but they were compiled using a different videojs core.
I tried multiple things to work around it but unfortunately nothing worked 100% in all of the full/light/esm/cdn variants :(

webpack/es-light.config.js
tsi: I thought `delete webpackConfig.entry` would be as explicit as it gets, what could be more explicit than that? :)

PR #766, opened 2024-12-17
fix: seek-thumbs positioning

tsi:
Fixes a regression caused by [this change](https://github.com/videojs/video.js/blame/19ca3f2ebe0ec7f0374fa3bc7df1457f9f185823/src/js/control-bar/progress-control/seek-bar.js#L52) in videojs.

Now, our hacky work-around is not needed anymore, and we can simply configure the player to behave nicely on mobile.


ehab-cl (approved):
💪

PR #768, opened 2024-12-19
me-17958: test if videos on chapters page are playing

ShayLevi:
Relevant task - https://cloudinary.atlassian.net/browse/ME-17958
This test is navigating to chapters page (chapters.html) and make sure that 3 videos element are playing (checking that the video is not paused).

### DIFF
--- a/test/e2e/specs/chaptersPage.spec.ts
+++ b/test/e2e/specs/chaptersPage.spec.ts
@@ -9,7 +9,7 @@ const link = getLinkByName(ExampleLinkName.Chapters);
 /**
  * Testing if videos on chapters are playing by checking that is pause return false.
  */
-vpTest.only(`Test if 3 videos on chapters page are playing as expected`, async ({ page, pomPages }) => {
+vpTest(`Test if 3 videos on chapters page are playing as expected`, async ({ page, pomPages }) => {
     await test.step('Navigate to chapters page by clicking on link', async () => {
         await pomPages.mainPage.clickLinkByName(link.name);
         await waitForPageToLoadWithTimeout(page, 5000);
--- a/test/e2e/specs/chaptersPage.spec.ts
+++ b/test/e2e/specs/chaptersPage.spec.ts
@@ -4,11 +4,8 @@ import { waitForPageToLoadWithTimeout } from '../src/helpers/waitForPageToLoadWi
 import { getLinkByName } from '../testData/pageLinksData';
 import { ExampleLinkName } from '../testData/ExampleLinkNames';
 
-// Link to chapters page
 const link = getLinkByName(ExampleLinkName.Chapters);
-/**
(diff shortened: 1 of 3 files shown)


refael-m (approved):

test/e2e/specs/chaptersPage.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { vpTest } from '../fixtures/vpTest';
+import { expect, test } from '@playwright/test';
+import { waitForPageToLoadWithTimeout } from '../src/helpers/waitForPageToLoadWithTimeout';
+import { getLinkByName } from '../testData/pageLinksData';
+import { ExampleLinkName } from '../testData/ExampleLinkNames';
+
+// Link to chapters page
```
ehab-cl, line 7: I think its already clear from the code, i will refrain adding comments when the code makes it clear ... this should also be the target, a self descriptive code 
```suggestion
```

test/e2e/specs/chaptersPage.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+const link = getLinkByName(ExampleLinkName.Chapters);
+/**
+ * Testing if videos on chapters are playing by checking that is pause return false.
```
ehab-cl, line 10: Same comment above

test/e2e/specs/chaptersPage.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+    await test.step('Validating that the second video is playing (in case isPause is false)', async () => {
+        expect(await pomPages.chaptersPage.chaptersConfigObjectVideoComponent.isPaused()).toEqual(false);
+    });
+    await test.step('Scroll until the third video element is visible', async () => {
+        await pomPages.chaptersPage.chapterAutoVttFileVideoComponent.locator.scrollIntoViewIfNeeded();
+    });
+    await test.step('Validating that the third video is playing (in case isPause is false)', async () => {
+        await expect(async () => {
+            expect(await pomPages.chaptersPage.chapterAutoVttFileVideoComponent.isPaused()).toEqual(false);
+        }).toPass({ intervals: [500], timeout: 3000 });
```
ehab-cl, line 29: why do you need specific timeout here?

ehab-cl (approved):
💪

test/e2e/specs/chaptersPage.spec.ts
ShayLevi: As we're scrolling to the element and only then it's playing we need to wait a bit to make sure the video is playing. It can take few seconds.

test/e2e/specs/chaptersPage.spec.ts
ShayLevi: Removed

PR #795, opened 2025-01-19
me-17978: tests if videos are playing on vast and vpaid page

ShayLevi:
Relevant task - https://cloudinary.atlassian.net/browse/ME-17978
This test is navigating to vast and vpaid page (vast-vpaid.html) and make sure that videos element are playing.
New optional parameter called `timeout` was added to `validateVideoIsPlaying` function because this page tests ads before videos and we need to wait minimum of 10 seconds until the ad finishes and the videos is starting. In all other cases `timeout` is set to default of 3 seconds.

### DIFF
--- a/test/e2e/specs/vastAndVpaidPage.spec.ts
+++ b/test/e2e/specs/vastAndVpaidPage.spec.ts
@@ -19,6 +19,6 @@ vpTest(`Test if 2 videos on vast and vpaid page are playing as expected`, async 
         await pomPages.vastAndVpaidPage.singleVideoWithAdsVideoComponent.validateVideoIsPlaying(true, 12000);
     });
     await test.step('Validating that playlist with ads video is playing', async () => {
-        await pomPages.vastAndVpaidPage.playlistWithAdsVideoComponent.validateVideoIsPlaying(true);
+        await pomPages.vastAndVpaidPage.playlistWithAdsVideoComponent.validateVideoIsPlaying(true, 12000);
     });
 });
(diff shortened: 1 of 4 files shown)


test/e2e/specs/vastAndVpaidPage.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    });
+    await test.step('Click on play button of single video with ads to play video', async () => {
+        return pomPages.vastAndVpaidPage.singleVideoWithAdsVideoComponent.clickPlay();
+    });
+    //Sending timeout of 12 seconds to wait until the ad finishes (10 sec) and the video will start
+    await test.step('Validating that single video with ads is playing', async () => {
+        await pomPages.vastAndVpaidPage.singleVideoWithAdsVideoComponent.validateVideoIsPlaying(true, 12000);
+    });
+    await test.step('Validating that playlist with ads video is playing', async () => {
+        await pomPages.vastAndVpaidPage.playlistWithAdsVideoComponent.validateVideoIsPlaying(true);
```
ehab-cl, line 22: I would also specify the timeout here, this is more explicit rather than what is currently implicit (The fact that this test is using the timeout of the previous step)... this in my eyes makes the test more readable

ehab-cl (approved):
💪

test/e2e/specs/vastAndVpaidPage.spec.ts
ShayLevi: Added

PR #816, opened 2025-02-12
fix: Non-text contrast (WCAG 1.4.11)

tsi:
This PR fixes an issue with the volume slider track color-contrast

Before:
image

After:
image

### DIFF
--- a/src/assets/styles/main.scss
+++ b/src/assets/styles/main.scss
@@ -255,6 +255,11 @@ $icon-font-path: '../icon-font' !default;
       &.vjs-volume-panel-horizontal {
         max-width: 8em;
       }
+
+      .vjs-slider {
+        background-color: color-mix(in srgb, var(--color-text) 10%, transparent);
+        box-shadow: 0 0 1px 1px color-mix(in srgb, var(--color-text) 80%, transparent) inset;
+      }
     }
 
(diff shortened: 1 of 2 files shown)


src/assets/styles/main.scss
```diff
@@ -255,6 +255,11 @@ $icon-font-path: '../icon-font' !default;
       &.vjs-volume-panel-horizontal {
         max-width: 8em;
       }
+
+      .vjs-slider {
+        background-color: color-mix(in srgb, var(--color-text) 10%, transparent);
```
kalifyaniv-cloudinary, line 260: we have both `--text-color` and `--color-text` ? what is the diff?

kalifyaniv-cloudinary (approved):

src/assets/styles/main.scss
tsi: Missed this, we actually have two system of color variables, an old one in JS we implemented before css-vars was a thing, then when css started supporting variable we started using that.
I would be very happy to move everything to CSS vars, it would improve the flow and reduce JS bundle size

tsi: Created https://cloudinary.atlassian.net/browse/ME-19139 :)

PR #817, opened 2025-02-12
fix: Contrast (Minimum) (WCAG SC 1.4.3)

tsi:
This PR fixes a few text color-contrast issues

Before:
image

After:
image

### DIFF
--- a/src/video-player.utils.js
+++ b/src/video-player.utils.js
@@ -108,8 +108,9 @@ export const overrideDefaultVideojsComponents = () => {
   const Player = videojs.getComponent('Player');
   let children = Player.prototype.options_.children;
 
-  // Add TitleBar as default
-  children.push('titleBar');
+  if (children.indexOf('titleBar') === -1) {
+    children.push('titleBar');
+  }
 
   const ControlBar = videojs.getComponent('ControlBar');
   if (ControlBar) {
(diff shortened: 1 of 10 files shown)


src/video-player.utils.js
```diff
@@ -108,8 +108,9 @@ export const overrideDefaultVideojsComponents = () => {
   const Player = videojs.getComponent('Player');
   let children = Player.prototype.options_.children;
 
-  // Add TitleBar as default
-  children.push('titleBar');
+  if (children.indexOf('titleBar') === -1) {
```
kalifyaniv-cloudinary, line 111: why we need this "if" ?

kalifyaniv-cloudinary (approved):

src/video-player.utils.js
tsi: I was playing with colors opacity, and got weird results visually, this is how I found out the title is being rendered twice :)
This `if` prevents that.

tsi: @ShayLevi could this be what's failing the test here? I couldn't find any reason why it would fail, the videos obviously play fine in the [deploy preview](https://deploy-preview-817--cld-video-player.netlify.app/docs/multiple-players)

ShayLevi: @tsi is it possible that the video element `id` was changed? I see we have failures on multiple players page and colors API page due to that.

PR #819, opened 2025-02-17
fix: add live stream error screen & auto refresh

jakub-roch:
Add live stream error screen when `type === 'live'` & auto refresh every 10 seconds

### DIFF
--- a/src/utils/get-analytics-player-options.js
+++ b/src/utils/get-analytics-player-options.js
@@ -109,6 +109,7 @@ export const getAnalyticsFromPlayerOptions = (playerOptions) => filterDefaultsAn
   width: playerOptions.width,
   withCredentials: playerOptions.withCredentials,
   debug: playerOptions.debug,
+  type: playerOptions.type,
 
   colors: playerOptions.colors && JSON.stringify(playerOptions.colors),
   controlBar: (JSON.stringify(playerOptions.controlBar) !== JSON.stringify(defaults.controlBar)) && JSON.stringify(playerOptions.controlBar),
--- a/src/video-player.js
+++ b/src/video-player.js
@@ -176,10 +176,7 @@ class VideoPlayer extends Utils.mixin(Eventable) {
 
     this.videojs.on(PLAYER_EVENT.PLAY, this._resetReTryVideoState);
     this.videojs.on(PLAYER_EVENT.CAN_PLAY_THROUGH, this._resetReTryVideoState);
-    this.videojs.on(PLAYER_EVENT.CLD_SOURCE_CHANGED, (...args) => {
-      this._onSourceChange(...args);
-      this.isLiveStream = args[1].source.resourceConfig().type === 'live';
-    });
+    this.videojs.on(PLAYER_EVENT.CLD_SOURCE_CHANGED, this._onSourceChange.bind(this));
 
     this.videojs.ready(() => {
       this._onReady();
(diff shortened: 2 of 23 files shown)


src/video-player.js
```diff
@@ -134,6 +135,28 @@ class VideoPlayer extends Utils.mixin(Eventable) {
...
+
+    tempVideo.onprogress = () => {
+      tempVideo.onerror = null;
+      tempVideo.oncanplaythrough = null;
+      this.source(liveStreamUrl);
+    };
+    tempVideo.onerror = retryLiveStreamLoad;
+
+    retryLiveStreamLoad();
+  }
```
tsi, line 158: @jakub-roch we already have a reTryVideo handler.
can we re-use that instead?
https://github.com/cloudinary/cloudinary-video-player/blob/master/src/video-player.js#L418

src/utils/get-analytics-player-options.js
```diff
@@ -109,6 +109,7 @@ export const getAnalyticsFromPlayerOptions = (playerOptions) => filterDefaultsAn
   width: playerOptions.width,
   withCredentials: playerOptions.withCredentials,
   debug: playerOptions.debug,
+  type: playerOptions.type,
```
tsi, line 112: 💪

src/video-player.js
```diff
@@ -50,11 +53,10 @@ class VideoPlayer extends Utils.mixin(Eventable) {
   constructor(elem, initOptions, ready) {
     super();
 
-    this.nbCalls = 0;
-
     this.videoElement = getResolveVideoElement(elem);
 
     this.options = extractOptions(this.videoElement, initOptions);
+    this.isLiveStream = this.options.videojsOptions.type === 'live';
```
tsi, line 59: Should we set this as part of player constructor? Imean, the initial source can be live but the same player can switch to other sources that are not.
[Here](https://github.com/cloudinary/cloudinary-video-player/blob/ME-19040-live-stream-error/src/video-player.js#L261) we do `source.resourceConfig().type === 'live'`, would that be better?

tsi (approved):
@jakub-roch looks much better.
I'm approving to unblock you but please see my comments

src/video-player.js
```diff
@@ -557,11 +580,10 @@ class VideoPlayer extends Utils.mixin(Eventable) {
       options.withCredentials = true;
     }
 
-    clearTimeout(this.reTryId);
-    this.nbCalls = 0;
+    this._resetReTryVideoState();
     const maxTries = this.videojs.options_.maxTries || 3;
     const videoReadyTimeout = this.videojs.options_.videoTimeout || 55000;
```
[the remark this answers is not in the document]
jakub-roch: Im not sure abut changing it - will check it to make sure we can replace it with `RETRY_DEFAULT_TIMEOUT`

src/video-player.js
```diff
@@ -415,26 +425,39 @@ class VideoPlayer extends Utils.mixin(Eventable) {
...
+      this.reTryVideoStateRetriesCount = 0;
     }
   }
 
-  isVideoReady() {
-    const s = this.videojs.readyState();
-    if (s >= (/iPad|iPhone|iPod/.test(navigator.userAgent) ? 1 : 4)) {
-      this.nbCalls = 0;
-      return true;
+  reloadVideoUntilAvailable(maxNumberOfCalls = Number.POSITIVE_INFINITY, timeout = RETRY_DEFAULT_TIMEOUT) {
```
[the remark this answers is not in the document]
jakub-roch: I dont see a reason why not - it makes sure that until user close tab / browser it will still re-try to load video
it just repeats fetch calls for video to see if its available

src/video-player.js
```diff
@@ -415,26 +425,39 @@ class VideoPlayer extends Utils.mixin(Eventable) {
...
+    if (typeof this.reloadVideoRetriesCount !== 'number') {
+      this.reloadVideoRetriesCount = 0;
     }
 
-    return false;
+    if (this.reloadVideoRetriesCount < maxNumberOfCalls) {
```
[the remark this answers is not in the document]
jakub-roch: nice catch, I missed it 👍

src/video-player.js
jakub-roch: good point, I added `isLiveStream` fn that is called every time to check current type 👍

jakub-roch: anyway its edge case that type of live will be changed but anyway good to handle it

PR #838, opened 2025-03-30
me-17994: test if video is playing on ESM components page

ShayLevi:
Relevant task - https://cloudinary.atlassian.net/browse/ME-17994
This test is navigating to ESM components page and make sure that video element is playing.
As it share common steps as `componentsPage.spec.ts` that was already implemented I created common spec function `testComponentsPageVideoIsPlaying` and using it on both specs.


tsi (approved):
✅

refael-m (approved):

PR #839, opened 2025-03-31
me-17995: test if video is playing on ESM display configuration page

ShayLevi:
Relevant task - https://cloudinary.atlassian.net/browse/ME-17995
This test is navigating to ESM display configuration page and make sure that video element is playing.
As it share common steps as `displayConfigurationsPage.spec.ts` that was already implemented I created common spec function `testDisplayConfigurationPageVideoIsPlaying` and using it on both specs.
Also added autoplay attribute to video element on ESM display configuration page to be the same as the equivalent base example page (https://cloudinary.github.io/cloudinary-video-player/ui-config.html)

### DIFF
--- a/docs/es-modules/ui-config.html
+++ b/docs/es-modules/ui-config.html
@@ -35,6 +35,7 @@
         crossorigin="anonymous"
         controls
         muted
+        autoplay
         playsinline
       ></video>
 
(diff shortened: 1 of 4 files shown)


tsi (approved):
👍

refael-m (approved):

PR #845, opened 2025-04-06
fix: visual search plugin accessibility

tsi:
This PR improves a few aspects of using the visual-search plugin with a keyboard.
- Button has a visible focus highlight
- Input and X are focusable only when visible
- Results marker are focusable and clickable

### DIFF
--- a/src/plugins/visual-search/components/SearchResults.js
+++ b/src/plugins/visual-search/components/SearchResults.js
@@ -1,51 +1,69 @@
 import videojs from 'video.js';
 
-export const SearchResults = (player) => {
+export const SearchResults = player => {
   const clearMarkers = () => {
     player.$$('.vjs-visual-search-marker').forEach(el => el.remove());
     player.$$('.vjs-visual-search-results-wrapper').forEach(el => el.remove());
-    
+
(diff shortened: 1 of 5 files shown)


src/plugins/visual-search/components/SearchResults.js
```diff
@@ -1,51 +1,69 @@
...
-      
+
       // Add click handler to jump to this time
       markerEl.addEventListener('click', () => {
         player.currentTime(start_time);
       });
+
+      // Add keyboard support
+      markerEl.addEventListener('keydown', e => {
+        if (e.key === 'Enter' || e.key === ' ') {
```
kalifyaniv-cloudinary, line 51: space?

kalifyaniv-cloudinary (approved):

src/plugins/visual-search/components/SearchResults.js
tsi: We could do `e.code === 'Space'` which is more semantically clear, but AFAIU for assistive technology devices that aren't standard keyboards, e.key will give better support because it represents the semantic _meaning_ of the input ("space") and not necessarily the space-key itself.

PR #862, opened 2025-05-12
feat: live streams analytics

jakub-roch:
Add new version of video analytics:
* use player adapter
* add live stream analytics

### DIFF
--- a/src/plugins/cloudinary-analytics/index.js
+++ b/src/plugins/cloudinary-analytics/index.js
@@ -40,7 +40,7 @@ class CloudinaryAnalytics {
   };
 
   getCloudinaryVideoPlayerAdapter = () => {
-    const createCldVPEventListener = (callback, eventName) => {
+    const createCldVPEventListener = (eventName, callback) => {
       this.player.on(eventName, callback);
       return () => {
         this.player.off(eventName, callback);
--- a/src/plugins/cloudinary-analytics/index.js
+++ b/src/plugins/cloudinary-analytics/index.js
@@ -1,4 +1,3 @@
-import videojs from 'video.js';
 import isPlainObject from 'lodash/isPlainObject';
 import { connectCloudinaryAnalytics } from 'cloudinary-video-analytics';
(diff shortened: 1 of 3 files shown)


src/plugins/cloudinary-analytics/index.js
```diff
@@ -22,45 +20,65 @@ class CloudinaryAnalytics {
...
+      onSuspend: (callback) => createCldVPEventListener('suspend', callback),
+      onTimeUpdate: (callback) => createCldVPEventListener('timeupdate', callback),
+      onVolumeChange: (callback) => createCldVPEventListener('volumechange', callback),
+      onWaiting: (callback) => createCldVPEventListener('waiting', callback),
+
+      getCurrentSrc: () => this.player.videoElement.currentSrc,
+      getCurrentTime: () => this.player.currentTime(),
+      getReadyState: () => this.player.videoElement.readyState,
+      getDuration: () => this.player.duration(),
+    };
```
kalifyaniv-cloudinary, line 78: does this mean we added more events to the "regular" tracking also?

src/plugins/cloudinary-analytics/index.js
```diff
@@ -22,45 +20,65 @@ class CloudinaryAnalytics {
...
   };
 
   init() {
-    // only for iOS there is problem with reporting events because videojs re-triggers events and stops native ones
-    if (this.shouldUseCustomEvents) {
-      this.connectCustomEvents();
-    }
-
```
kalifyaniv-cloudinary, line 63: why did we remove this?

kalifyaniv-cloudinary (approved):
Please check my second comment before merging...

src/plugins/cloudinary-analytics/index.js
jakub-roch: for now - no, I just created mapping for all events, it doesn't mean we are listening for all of them, it just means that we can start listening whenever we want to ( thats schema of required props as player adapter )

PR #904, opened 2025-08-25
feat: source switcher

jakub-roch:
Source switcher

image


src/components/source-switcher-button/source-switcher-button.js
```diff
@@ -0,0 +1,137 @@
...
+
+  createItems() {
+    if (!Array.isArray(this._items) || this._items.length === 0) {
+      const empty = new MenuItem(this.player_, {
+        label: this._emptyLabel || 'No sources',
+        selectable: false
+      });
+      empty.addClass('vjs-source-switcher-empty');
+      empty.disable();
+      return [empty];
```
kalifyaniv-cloudinary, line 38: why do we have an "empty" state?

kalifyaniv-cloudinary (approved):

jakub-roch:
> @jakub-roch how will this play together with a `publicId` configuration? or with a player using the `source(...)` method? will I be able to pass an array to `source()`?

@tsi for now we dont really care about any incorrect usage (like blocking other methods when source is being used), user can provide publicId but it will be ignored and first item of `videoSources` list will be used

tsi:
> dont really care about any incorrect usage

@jakub-roch why using the `source()` method to switch videos is incorrect? how about using these alternate-sources in a playlist?

Also, please add this to internal analytics

src/components/source-switcher-button/source-switcher-button.js
```diff
@@ -0,0 +1,135 @@
...
+        const payload = { index, value, label };
+        if (this._onSelected) this._onSelected(payload, this.player_);
+        this.player_.trigger('sourceswitcher:change', payload);
+      });
+
+      return item;
+    });
+  }
+
+  setItems(items, keepSelection = true) {
```
tsi, line 65: Why `keepSelection` defaults to false if its only usage needs `false`? looks like this can be simplified

docs/index.html
```diff
@@ -76,6 +76,7 @@ <h3 class="mt-4">Some code examples:</h3>
       <li><a href="./seek-thumbs.html">Seek Thumbnails</a></li>
       <li><a href="./share-plugin.html">Share &amp; Download</a></li>
       <li><a href="./shoppable.html">Shoppable Videos</a></li>
+      <li><a href="./source-switcher.html">Source switcher</a></li>
```
tsi, line 79: Should we also add an ESM example page? I'm ok with not having everything doubled, just that this is what we did with other examples

src/validators/validators.js
```diff
@@ -37,6 +37,7 @@ export const playerValidators = {
   },
   playerOptions: {
     debug: validator.isBoolean,
+    videoSources: validator.isArray,
```
tsi, line 40: Move to `sourceValidators`?

src/utils/get-analytics-player-options.js
```diff
@@ -117,6 +117,7 @@ export const getAnalyticsFromPlayerOptions = (playerOptions) => filterDefaultsAn
   withCredentials: playerOptions.withCredentials,
   debug: playerOptions.debug,
   type: playerOptions.type,
+  videoSources: !!playerOptions.videoSources,
```
tsi, line 120: Move to `getSourceOptions`?

tsi (approved):
Looks and works great! 💪 
https://codepen.io/tsi/pen/qEOyQGr

docs/index.html
jakub-roch: we dont have yet E2E test for this page (just mocked for now) but I believe ESM should be added only if there is a reason for that (otherwise we are doubling everything), for now I will keep just one page but we need to decide about some rules here (will schedule meeting for us after gathering)

PR #906, opened 2025-08-25
fix(VIDEO-19962): base64 encode public id for video metadata

kjarmicki-cloudinary:
We need to encode public ID because it can contain URL-unfriendly things like slashes


tsi (approved):
💪

PR #1024, opened 2026-04-06
fix: stabilize _s query params for better browser caching

tsi:
## Summary

Use a stable `_s=vp` marker for usage-reporting query strings so URLs cache more predictably across player versions, and avoid merging the video source’s `queryParams` onto **raw absolute** poster URLs so external poster hrefs stay exactly as provided.

## Changes

- Set `_s` to `vp` (instead of `vp-<package version>`) when `allowUsageReport` is on in Cloudinary `buildSource` and in `fetch-config` applet requests.
- In `VideoSource.poster()`, apply `this.queryParams()` only when the poster target is not a raw URL, preserving untouched raw poster URLs while still appending params for Cloudinary-derived posters.
- Add unit tests for raw poster hrefs vs Cloudinary-built poster URLs.


src/plugins/cloudinary/index.js
```diff
@@ -207,7 +207,7 @@ class CloudinaryContext {
         { hasUserPosterOptions: hasUserPosterOptions || null }
       );
 
-      options.queryParams = Object.assign(options.queryParams || {}, options.allowUsageReport ? { _s: `vp-${VERSION}` } : {});
+      options.queryParams = Object.assign(options.queryParams || {}, options.allowUsageReport ? { _s: 'vp' } : {});
```
kalifyaniv-cloudinary, line 210: because we report the version in the internal analytics call? 
why do we remove the version?

kalifyaniv-cloudinary (approved):

PR #1042, opened 2026-05-03
fix: improve error message display

tsi:
## Summary
Improves the Cloudinary Video Player error overlay so errors render as a centered, themed message instead of the previous flat gray block. Adds a dedicated title, message area, and retry action wired to the existing refresh event.

## Changes
- Adds structured error display content with a `Something went wrong` heading and localized error message body.
- Adds a `Try Again` action when the error message does not already provide a `.cld-error-refresh` link.
- Updates error overlay styling to use the player theme colors, centered layout, a muted poster backdrop, and a compact error panel.

## Visuals
UI styling changed for player error states. Please attach before/after screenshots or verify the error overlay visually before review.

## Before:
image

## After:
image

### DIFF
--- a/src/components/error-display/error-display.js
+++ b/src/components/error-display/error-display.js
@@ -9,12 +9,18 @@ class CldErrorDisplay extends ErrorDisplay {
     if (!error) return '';
 
     const wrapper = videojs.dom.createEl('div', { className: 'cld-error-display-content' });
-    const msg = videojs.dom.createEl('span', { className: 'cld-error-message' });
+    const header = videojs.dom.createEl('div', { className: 'cld-error-header' });
+    const title = videojs.dom.createEl('span', { className: 'cld-error-title' }, {}, this.localize('Something went wrong'));
+    const msg = videojs.dom.createEl('div', { className: 'cld-error-message' });
+
+    header.appendChild(title);
+
     if (error.isHtml) {
       msg.innerHTML = this.localize(error.message);
     } else {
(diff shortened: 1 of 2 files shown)


src/components/error-display/error-display.js
```diff
@@ -9,12 +9,18 @@ class CldErrorDisplay extends ErrorDisplay {
     if (!error) return '';
 
     const wrapper = videojs.dom.createEl('div', { className: 'cld-error-display-content' });
-    const msg = videojs.dom.createEl('span', { className: 'cld-error-message' });
+    const header = videojs.dom.createEl('div', { className: 'cld-error-header' });
+    const title = videojs.dom.createEl('span', { className: 'cld-error-title' }, {}, this.localize('Something went wrong'));
```
adimiz1, line 13: Something went wrong - should be hardcoded?

adimiz1, line 13: Is cld-error-title needed here? I could not find it in code

adimiz1 (approved):
No test added for new header/refresh-conditional behavior, maybe it's worth adding one?

src/components/error-display/error-display.js
tsi: I wanted some generic title, so a default one must be hard-coded, but we can allow it to be configured I guess

tsi: wdym? we give this class to the "Something went wrong" span

PR #1045, opened 2026-05-07
fix: [VIDEO-20833] track lazy bundle analytics

tsi:
## Summary

Lazy entry paths (`videoPlayer`, `player`, and legacy `Cloudinary.new()` wrappers) now merge `_internalAnalyticsMetadata` with `playerBundle: 'lazy'` so internal analytics can tell lazy-loaded core apart from eager bundles.

## Changes

- Introduce a shared `mergeOpts` helper that shallow-merges option objects and merges `_internalAnalyticsMetadata` from each layer, then sets `playerBundle` to `'lazy'`.
- Pass merged options into deferred `createVideoPlayer` and async `createPlayerWithConfig` flows.
- Legacy `cloudinaryVideoPlayerLegacyConfig` uses the same helper so instance config and per-call options merge consistently (instance + call) while still tagging lazy analytics.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,10 +3,17 @@ import { createAsyncPlayer, createMultiplePlayers, createMultipleSync, setupClou
 import { createDeferredPlayer } from './utils/deferred-player';
 
 const importCore = () => import(/* webpackChunkName: "cld-player-core" */ './video-player.js');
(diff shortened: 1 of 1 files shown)


adimiz1 (approved):
Maybe add a test?

PR #1066, opened 2026-07-16
feat(types): add event, videojs handle, and default export typings

kdzialdowski-cloudinary:
The shipped `VideoPlayer` typings omit several members that exist at runtime, forcing consumers to augment the module locally. This adds them to `types/cld-video-player.d.ts`: the `on`/`one`/`off`/`trigger` event methods (wired up by `setupEventMethods`), the underlying `videojs` player handle (e.g. `.videojs.el()`), and the `Player` reference plus `eventData` payload attached to forwarded extended events. It also adds the missing `export default` for the cloudinary global, which exists at runtime (`src/index.js`) but was absent from the declarations.

Extended events get per-event overloads so their payloads are typed precisely: `percentsplayed` → `eventData.percent`, `timeplayed` → `eventData.time`, `seek` → `eventData.{seekStart,seekEnd}`. A trailing `string` overload covers all other video.js/cloudinary events.

These gaps surfaced when bumping the package in a consuming project (gallery-widget), where the fix was a local `declare module` augment — contributing the typings upstream removes the need for that. The type surface stays self-contained (no video.js type dependency), and `types/cld-video-player-tests.ts` is extended to validate the new members. Verified that gallery-widget typechecks clean with its local augment removed.

### DIFF
--- a/types/cld-video-player-tests.ts
+++ b/types/cld-video-player-tests.ts
@@ -27,3 +27,28 @@ async () => {
     profile: 'default'
   });
 };
+
+// Event methods and the video.js handle.
+player.on('ready', () => {});
+player.on('play', (event) => {
+  const emitter: VideoPlayer = event.Player;
+  emitter.pause();
+});
+player.on('percentsplayed', (event) => {
+  const percent: number = event.eventData.percent;
+  void percent;
+});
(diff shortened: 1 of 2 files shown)


tsi (approved):
Looking good, thanks!!

PR #1074, opened 2026-07-29
fix(ui): show full playback rate values in the rates menu

danieldanieli:
video.js caps the playback rate menu at a fixed 4em. Once the menu item padding and the selected-item checkmark gutter are subtracted, rates like 1.75x no longer fit and get ellipsized to "1...".

The menu content is absolutely positioned, so it - not the menu box - is the element that has to grow: size it to max-content, with min-width 100% so short rate lists keep the original popup width. Also widen the control itself to 3em, since the 1.75x label measures 36.1px against the stock 2.5em (36px) button and crowds the neighbouring control.

### DIFF
--- a/src/assets/styles/main.scss
+++ b/src/assets/styles/main.scss
@@ -384,9 +384,22 @@ $icon-font-path: '../icon-font' !default;
     }
   }
 
-  .vjs-playback-rate-value {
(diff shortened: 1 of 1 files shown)


tsi (approved):
🔥

PR #1076, opened 2026-07-29
feat(config): default preload to metadata

danieldanieli:
Align the player default with standard browser behavior and cut the initial network cost: `auto` lets the browser pull media data for a video the user may never play, while `metadata` fetches only what is needed to report duration and dimensions.

Updated in both places the default is declared - defaults.js (runtime) and configSchema.json, which is copied into the published bundles as the documented option reference.

Note this governs the native <video> element, so it applies to progressive sources; hls.js drives its own loading for adaptive streaming, and autoplay still fetches media data either way.

### DIFF
--- a/src/config/configSchema.json
+++ b/src/config/configSchema.json
@@ -66,7 +66,7 @@
     "preload": {
       "type": "string",
       "enum": ["auto", "metadata", "none"],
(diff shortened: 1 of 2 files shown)


tsi (approved):
👍

PR #1086, opened 2026-09-14
fix: allow textTracks redefinition on repeated HLS source-set

adimiz1:
## Summary
- Customer reported `TypeError: Cannot redefine property: textTracks` and dead playback when a second HLS source is set on the same player/element — e.g. `hls/h265` returns 400 and the app falls back to `hls/h264` on the same instance.
- Root cause: `Html5HlsJS()` calls `Object.defineProperty(el, 'textTracks', {...})` on every source-set, but `configurable` was never set, so it defaults to `false`. The second call throws before `hls.attachMedia()`/`hls.loadSource()` ever run, so the fallback source never attaches.
- Fix: add `configurable: true` to that descriptor.

## Trade-off
Adding `configurable: true` removes a fail-fast guard. If some future refactor ever causes `Html5HlsJS()` to run twice on the same `el` **without** going through video.js's normal dispose-first lifecycle (`Tech.prototype.setSource` → `disposeSourceHandler()` → `hls.destroy()` → new source handler), it will now silently reconfigure the property instead of throwing where it would previously have been caught in testing/monitoring. Verified via `video.js` core that the current, real call pattern always disposes the old source handler and reuses the same `tech` object before creating the new one — so the getter closure is functionally identical before and after redefinition, and this trade-off does not present a risk in the actual call path today. It's a latent risk for a *different*, currently-hypothetical bug class only.

## Verification
- Reproduced live in-browser against the demo page (`docs/adaptive-streaming.html`): confirmed the exact reported stack trace (`Html5HlsJS` → `handleSource` → `Tech.setSource` → `Player.src_`) with the fix reverted, and confirmed clean playback with the fix applied.
- Added `test/unit/videojs-contrib-hlsjs-texttracks.test.js` as a permanent regression test, exercising the real module via a captured `HlsSourceHandler` reference (mocking only `hls.js`, irrelevant to this bug).

## Out of scope (flagged separately, not fixed here)
- `el.addEventListener('error', ...)` is re-added on every `Html5HlsJS()` call but never removed by `dispose()` — a pre-existing stale-listener leak on repeated source-sets, independent of this bug.
- Local e2e (`test/adaptive-streaming.test.js` via jest-puppeteer) is currently broken in this checkout for unrelated reasons (`jest-environment-puppeteer` export-shape mismatch, `page.waitFor` removed in the installed Puppeteer version) — confirmed the *pre-existing* tests in that file also fail in isolation, unrelated to this change.

## Test plan
- [x] Reproduced the bug locally, verified the exact reported error
- [x] Applied fix, confirmed second HLS source-set no longer throws, live in browser
- [x] Added unit regression test (`test/unit/videojs-contrib-hlsjs-texttracks.test.js`)
- [ ] CI unit test run (local vitest run blocked by an unrelated vite/vitest version mismatch in this checkout)


tsi (approved):
👍

PR #1088, opened 2026-09-15
fix(captions): size cue text by width on portrait players and let the box follow its content

tsi:
## Summary

Captions on portrait (9:16) players rendered with a huge font, got cut off at the bottom of the player and jumped between cues; any font-size override made the text float in the middle instead. Reported for Tudor's training video ([VIDEO-21222](https://cloudinary.atlassian.net/browse/VIDEO-21222)).

## Root cause

- vtt.js sets the cue font to **5% of the display height**. At 390px width that is ~11px for 16:9 but ~35px for 9:16.
- video.js 8 measures the cue and pixel-fixes its box (`top/height/width`) in `WebVTT.processCues` and only **afterwards** adds the `vjs-text-track-cue` class. Every rule we key on that class (Inter, 700, padding, inline-block) re-wraps the text after the box is frozen, so it spills out of the bottom.
- The same freeze means a `font-size` override (page CSS or `styledTextTracks.fontSize`) shrinks the text inside a still-tall box: the text floats and moves with each cue's line count.

## Changes

- `text-tracks.scss`: `container-type: size` on the caption display and `font-size: min(5cqh, 3.5cqw)` on cues. Landscape is unchanged (the `5cqh` term is vtt.js's own size); portrait is capped at 3.5% of the width.
- `text-tracks.scss` / `styled-text-tracks.scss`: `height: auto; top: auto` on bottom-gravity cues (`bottom: auto; height: auto` for top gravity) so the box follows its content, the same trick video.js uses for its own `fontPercent` setting. Center gravity untouched.
- `docs/subtitles-and-captions.html`: new "Portrait (9:16) player" section.
- New e2e: plays the portrait demo and asserts over 3+ cues that the text stays inside the caption display and the font is at most 4% of the player width. It fails on master with `cue "…" spills below the caption display`.

## Notes

- VTT/SRT cues with an explicit `line:` position from the top keep their computed bottom edge and grow upward; only visible if the text re-wraps. Accepted trade-off.
- Browsers without container-query units (Chrome < 105, Safari < 16, Firefox < 110) ignore the `min()` declaration and keep today's size, but still get the content-following box.
- `styledTextTracks.fontSize` keeps working; percentages are now relative to the responsive base.

## Testing

- `npx playwright test … subtitlesAndCaptionsPage.spec.ts`: 2/2 (existing + new)
- `npm run test:unit`: 129 passed
- `npm run lint`: clean
- Manually checked the karaoke demo (top gravity, 200% font) grows downward, and that the caption follows the control bar when it shows/hides.

## Follow-ups

- Embed page (`player.cloudinary.com`) sets `font-family: Inter` with no fallback stack, so iOS renders Times until the webfont loads. Separate repo.
- `specs/portrait-captions/` (spec + decision log) is here for review and will be removed before merge.


specs/portrait-captions/log.md
```diff
@@ -0,0 +1,70 @@
+# Decision log — Portrait captions fix (VIDEO-21222)
```
adimiz1: Dont forget to remove before merge

specs/portrait-captions/spec.md
```diff
@@ -0,0 +1,80 @@
+# Feature Specification: Captions on portrait players
```
adimiz1: Dont forget to remove before merge

adimiz1:
Non-blocking review notes:

1. **Native fullscreen scope check** — the fix targets the JS overlay (`.vjs-text-track-display` / `.vjs-text-track-cue`). iOS native fullscreen renders captions through `::cue` / `::-webkit-media-text-track-display` instead,  a separate pipeline, untouched here. Was the original repro (Tudor's video) in native fullscreen or inline/JS-controls view? If native fullscreen, this may not fully resolve the reported issue — worth confirming before closing VIDEO-21222.

2. **`videojs-default` theme behavior change**, `font-size: min(5cqh, 3.5cqw) !important` now applies unconditionally, including the `videojs-default` theme, whose purpose is matching vanilla video.js styling. Portrait/narrow players on that theme will now diverge from stock video.js sizing. Seems like a deliberate, reasonable call, but worth a one-line mention in the PR description so it's not a surprise later.

Smaller/non-essential:
- `text-tracks.scss`: the `&.vjs-user-inactive.vjs-playing .vjs-text-track-display` branch duplicated into the base selector block is redundant, the base `.vjs-text-track-display` selector already covers it. Could just drop that state from the old override block instead.
- `styledTextTracks.box` (custom width/height on the display) + the new `container-type: size` combo isn't explicitly covered by the manual testing notes — probably fine (explicit dims win) but worth a quick manual check.
- The new e2e polls up to 45s for 3 distinct live cues from the `prod` cloud asset — some flake/CI-time risk, and couples the docs demo + test to that asset's captions never changing.

None of these block merge in my view — just flagging for awareness / possible follow-up.

adimiz1 (approved):

